PR #339, opened 2023-11-10
fix: decrease <aside> height to prevent footer overlap

cellomatt:
### Description
At screen widths >= to 1280px, the `<aside>` component which renders the "On This Page" sidebar has a height which overlaps the footer. This means that the "Visit website" and "Contact us" links are unusable, because the rollover is not detected.

This PR increases the height calc from 95px to 200px (thus decreasing overall component height) to leave enough clearance for the links to remain clickable.


cjbell (approved):
💯

PR #346, opened 2023-11-29
feat: add reference to actors under messages

cjbell:
### Description

Add reference to actors under messages.

### Tasks
KNO-4756


francoborr (approved):
💯

PR #349, opened 2023-12-04
feat: add in-app feed response filter docs

cjbell:
### Description

Adds new docs under "Integrations > In-app > Knock" for customizing the API response from the feed endpoint.

Note: I ran `format.write` on all our contents, hence the noisy diff. Sorry! 

### Tasks
[KNO-4792](https://linear.app/knock/issue/KNO-4792/[docs]-api-response-filters-for-in-app-feeds)


connorlindsey (approved):
Looks good to me. Thanks for updating the format check

PR #351, opened 2023-12-13
fix: update getting started section under supported email providers

rachael-t:
### Description

Updating the "Getting Started" section for the different supported providers under integrations > email to all point users to "Integrations > Channels" from the dashboard.

Note that Mailtrap had already been updated.


cjbell (approved):
💯

PR #353, opened 2023-12-14
add callout for combined prefs

JEverhart383:
### Description

I updated the `Remember:...` section of text below [this code sample](https://docs.knock.app/concepts/preferences#combining-preferences) to be a callout. This change aligns it with the other uses of callouts the on concepts pages and better highlights an important caveat for the customer. 

### Screenshots
Before
Screenshot 2023-12-14 at 3 29 00 PM
After
Screenshot 2023-12-14 at 3 29 22 PM


cjbell (approved):
💯

PR #354, opened 2023-12-15
fix: change named to default import

JEverhart383:
### Description

Updates this code sample to use `Knock` as a default export instead of a named export. Checked these places:
https://github.com/knocklabs/knock-client-js/blob/main/src/index.ts
https://github.com/knocklabs/knock-node-example-app/blob/main/app/users/components/NotificationPreferencesModal.tsx

### Screenshots

Before
Screenshot 2023-12-15 at 8 50 12 AM

After
Screenshot 2023-12-15 at 8 51 01 AM


cjbell (approved):
💯

PR #356, opened 2023-12-15
fix: add default action_url behavior

JEverhart383:
### Description

Adds a recipe describing the default behavior of the `action_url` on the `KnockFeedProvider`. I put it first since its the most common scenario.

### Screenshots

Screenshot 2023-12-15 at 2 59 59 PM


cjbell (approved):
💯

PR #358, opened 2023-12-18
chore: add payload example for Segment Track events

cellomatt:
This PR makes 2 small clarifications to our Segment Extension documentation:

1. Add backticks to `track` to clarify that it is a Segment `event` type
2. Add an example event payload so that customers can see what Knock is sending to Segment

**Question:** Should we be using `<a target="_blank">` for external links, or should we be leaving these in pure MD even though it will direct users away from our site? I noticed that we are not currently preventing navigation away from the Knock docs with the link to Segment at the top of this particular page ([line 9](https://github.com/knocklabs/docs/blob/38d6a33fbae01bbb8ab95ea2657d8a2b3112cf06/content/integrations/extensions/segment.mdx#L9))

image

### DIFF
--- a/content/integrations/extensions/segment.mdx
+++ b/content/integrations/extensions/segment.mdx
@@ -46,7 +46,7 @@ Knock uses the Segment HTTP Source to send events to your Segment account. Each 
 
 ## Events sent to Segment
 
-When connected, Knock will forward the following track events to Segment.
+When connected, Knock will forward the following `track` events to Segment.
 
 <Table
   headers={["Event", "Description"]}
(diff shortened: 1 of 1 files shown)


content/integrations/extensions/segment.mdx
```diff
@@ -94,3 +94,36 @@ When connected, Knock will forward the following track events to Segment.
     ],
   ]}
 />
+
+Knock uses the Segment [Track spec](https://segment.com/docs/connections/spec/track/) for the event data we pass to Segment. You can find an example payload of what you can expect from Knock events coming into Segment below.
```
samseely, line 98: @cellomatt let's put an H2 above this called "Event schema" just so it shows up in table of contents of this page 👍

samseely (approved):
@cellomatt looks great! Had one comment on adding in an H2 but otherwise we're good to go. You can just keep the links in MD as we don't use target="_blank" in docs today

PR #360, opened 2023-12-19
chore: add Pigeon conversion instructions link to cert-based APNS setup instructions

cellomatt:
This PR adds documentation from Pigeon to inform customers of a required conversion that is necessary before they supply their APNS certificate and private key to the channel configuration in their dashboard. Needs review for content accuracy.


cjbell (approved):
💯

PR #371, opened 2024-01-04
chore: add timezone information to user concept page and API reference

cellomatt:
In response to user feedback about lacking documentation of the `timezone` parameter for User objects, this PR updates both the concept page for Users (/concepts/users) and the API reference for identifying a user (/reference#identify-user) in order to clarify its usage.

### DIFF
--- a/content/reference.mdx
+++ b/content/reference.mdx
@@ -616,10 +616,8 @@ Identifying a user will create or update a user in Knock, merging the properties
     type="string (optional)"
     description={
       <span>
-        The timezone of the user. Must be a valid 
-        <a href="https://en.wikipedia.org/wiki/List_of_tz_database_time_zones">tz database time zone string</a>
-        . Used for 
-        <a href="/concepts/schedules#scheduling-workflows-with-recurring-schedules-for-recipients">recurring schedules</a>.
+        The timezone of the user. Must be a valid <a href="https://en.wikipedia.org/wiki/List_of_tz_database_time_zones">tz database time zone string</a>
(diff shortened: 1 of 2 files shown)


content/reference.mdx
```diff
@@ -611,6 +611,19 @@ Identifying a user will create or update a user in Knock, merging the properties
...
+    name="timezone"
+    type="string (optional)"
+    description={
+      <>
+        <span>The timezone of the user. Must be a valid </span>
+        <a href="https://en.wikipedia.org/wiki/List_of_tz_database_time_zones">tz database time zone string</a>
+        <span>. Used for </span>
+        <a href="/concepts/schedules#scheduling-workflows-with-recurring-schedules-for-recipients">recurring schedules</a>
+        <span>.</span>
+      </>
```
cjbell, line 624:
```suggestion
      <span>
        The timezone of the user. Must be a valid 
        <a href="https://en.wikipedia.org/wiki/List_of_tz_database_time_zones">tz database time zone string</a>
        . Used for 
        <a href="/concepts/schedules#scheduling-workflows-with-recurring-schedules-for-recipients">recurring schedules</a>.
      </span>
```

cjbell (approved):
one suggestion

content/reference.mdx
cellomatt: I had copied this from the attribute above and also thought the syntax was odd, but now that I've committed it I see why the other example did that: it seems like when it's rendered, the `<span>` block doesn't respect the white space before the `<a>` tags when they are on new lines. I just pushed a new change that combines approaches, are we ok with inline `<a>` tags or should we go the original route? I think it's more readable inline.

PR #372, opened 2024-01-04
fix: update mistyped field to resolve JSX error

cellomatt:
A `name` field on an attribute in the API reference is mis-labeled as a `type`, resulting in a JSX error for two fields with the same name. This PR resolves the typo.

### DIFF
--- a/content/reference.mdx
+++ b/content/reference.mdx
@@ -3809,7 +3809,7 @@ Returns a paginated list of messages for an object. Will return newest messages 
     description="Limits the results to only items related to any of the provided categories"
   />
   <Attribute
-    type="JSON string (optional)"
+    name="JSON string (optional)"
     type="object (optional)"
     description="Limits the results to only items that were generated with the given data"
   />


content/reference.mdx
```diff
@@ -3809,7 +3809,7 @@ Returns a paginated list of messages for an object. Will return newest messages
     description="Limits the results to only items related to any of the provided categories"
   />
   <Attribute
-    type="JSON string (optional)"
+    name="JSON string (optional)"
```
connorlindsey, line 3812: Nice catch! It looks like the name should be `trigger_data` based on this PR: https://github.com/knocklabs/docs/commit/4472303908078551f0f46059e19053d893df7474

connorlindsey (changes requested):

content/reference.mdx
cellomatt: oh amazing, thanks for the context!

connorlindsey (approved):
🎉

PR #377, opened 2024-01-17
chore: update signUserToken example usage

connorlindsey:
### Description
Update docs so they're compatible with the latest Node SDK changes


thomaswhyyou (approved):
I'm guessing we merge this after the new version is published. 🙏

PR #379, opened 2024-01-22
Chore: Add callout regarding transferring email templates

rachael-t:
### Description

Adds a callout in the email layout documentation to remind users that `<head>` and `<body>` are included in the **layout** and should not be added to the **template** when transferring a template from an outside provider. Adding this to the documentation as this has come up twice in the last month when helping customers.

### DIFF
--- a/content/integrations/email/layouts.mdx
+++ b/content/integrations/email/layouts.mdx
@@ -236,3 +236,23 @@ Keep in mind that to override any of the base component style properties listed 
   margin-bottom: 0;
 }
 ```
+
+<br></br>
+
+<Callout
+  emoji="🚨"
+  text={
+    <>
+      <span className="font-bold">Reminder: </span>The email layout contains the{" "}
+      <code style={{ fontWeight: 600, backgroundColor: "inherit" }}>head</code>{" "}
+      and{" "}
+      <code style={{ fontWeight: 600, backgroundColor: "inherit" }}>body</code>{" "}
+      tags. If you are transferring an email template from an outside source,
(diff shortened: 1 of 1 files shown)


rachael-t:
Hey @cellomatt! Drafted up some copy we can add to the documentation per our pairing last week and customer feedback. I wanted to get *something* up, but I am *very* open to feedback on wording, location within the document, etc. 😅  Let me know your thoughts!

content/integrations/email/layouts.mdx
```diff
@@ -236,3 +236,23 @@ Keep in mind that to override any of the base component style properties listed
   margin-bottom: 0;
 }
 ```
+
+<br></br>
```
rachael-t, line 240: I don't feel like this is best practice, but without it, the callout was tightly against the code block above and didn't look great. We can (and should) remove this if we decide to adjust the formatting of the info below (i.e., move it elsewhere, change from a callout to just text, etc.).

cellomatt:
@rachael-t just adding the suggested copy edits here rather than in a code suggestion since you may want to change it up a bit more and that will affect all the `<span>` tags, etc. 😛 :

>Note: Since each Knock email layout contains `<head>` and `<body>` tags, these tags should not be included in your workflow's email template. If you are transferring email content from an outside source, you may need to make adjustments to remove any `<head>` and `<body>` tags prior to saving your email template within a Knock workflow.

cellomatt (approved):
lgtm! 🚀

connorlindsey (approved):

PR #380, opened 2024-01-23
chore: expand react migration guide

JEverhart383:
### Description

This PR makes some minor tweaks to the `@knocklabs/react` migration guide:

- call out provider API changes and CSS import updates
- make before/after code samples separate for readability
- use code samples from the feed implementation guide as examples for continuity between docs


connorlindsey (approved):
Looks awesome, thanks for adding more detail here!

PR #382, opened 2024-01-23
chore: update react native docs

connorlindsey:
### Description

Updates docs to use KnockProvider and KnockFeedProvider. Merge after we deploy a new version of the RN package.

Also fixes an issue in the react reference where we were still showing an example with the old KnockFeedProvider props


JEverhart383 (approved):
Updated one code sample, otherwise LGTM

PR #385, opened 2024-01-31
chore: add CI/CD CLI guide

cellomatt:
This PR adds a new guide on using the Knock CLI in your CI/CD pipeline. It also adds sidebar navigation for that guide, and a link from the main Knock CLI page back to this guide.

### DIFF
--- a/content/guides/cicd-with-the-knock-cli.mdx
+++ b/content/guides/cicd-with-the-knock-cli.mdx
@@ -73,21 +73,21 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - name: Checkout Repository
-      uses: actions/checkout@v4
+      - name: Checkout Repository
+        uses: actions/checkout@v4
 
-    - name: Set up Node.js
-      uses: actions/setup-node@v3
-      with:
-        node-version: '20.9.0' # Set Node.js version to ^16.14.0
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: "20.9.0" # Set Node.js version to ^16.14.0
 
-    - name: Install Knock CLI
-      run: npm install -g @knocklabs/cli
+      - name: Install Knock CLI
+        run: npm install -g @knocklabs/cli
 
-    - name: Promote Workflow to Staging
-      env:
-        KNOCK_SERVICE_TOKEN: ${{ secrets.KNOCK_SERVICE_TOKEN }} # Ensure you've set KNOCK_SERVICE_TOKEN in your GitHub secrets
-      run: knock commit promote --to=staging --service-token=$KNOCK_SERVICE_TOKEN
+      - name: Promote Workflow to Staging
+        env:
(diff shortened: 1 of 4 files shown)


kylemcd (approved):

content/guides/cicd-with-the-knock-cli.mdx
```diff
@@ -0,0 +1,129 @@
...
+section: Guides
+---
+
+With the [Knock CLI](/developer-tools/knock-cli), you can add Knock directly into your existing CI/CD pipeline in order to automate the promotion of changes to Workflows, Layouts, and Translations between [environments](/concepts/environments) during testing and deployment.
+
+This guide assumes that you have [created](/concepts/environments#create-additional-environments) a “Staging” environment in Knock in addition to the Development and Production environments that are provided by default. It also assumes that you have [installed the Knock CLI](/cli#installation) on your local machine.
+
+## Local development
+
+With the Knock CLI, you can pull resources from the Knock dashboard and develop them locally using the [`knock workflow pull --all`](/cli#workflow-pull), [`knock layout pull --all`](/cli#email-layout-pull), and [`knock translation pull --all`](/cli#translation-pull) commands. You can also work with specific resources by name rather than using the `--all` flag.
```
JEverhart383, line 24: Nit, but 'work with them by key' instead of 'name'

meryldakin (approved):
👏

data/sidebar.ts
```diff
@@ -218,6 +218,10 @@ const sidebarContent: SidebarSection[] = [
         slug: "/customizing-react-in-app-feed-components",
         title: "Customizing the React in-app feed components",
       },
+      {
+        slug: "/cicd-with-the-knock-cli",
```
samseely, line 222: @cellomatt only question I have is whether this should live under "Developer tools" section of sidebar. I tend to think of guides as walkthroughs for specific use cases and notification use cases (alerts, filtering in-app). 

In this case working with CI/CD feels like part of our developer workflow. cc @cjbell as he may have a POV on this.

samseely (approved):

data/sidebar.ts
cellomatt: Good question! I guess I was thinking about this as the CLI being the tool, this guide being the walkthrough of a use case. 

Wherever we put this, I should also go back and add a link to it from the Knock CLI tool page. I'll do that when I add some of the other CLI reference updates we've discussed.

cellomatt: @samseely I took another look at the other things we have under Guides right now and I see what you mean, they're more like _product_ use cases of things you might build with Knock tooling.

Maybe this content should just be added to the main [Knock CLI](https://docs.knock.app/developer-tools/knock-cli) page, as additional headers?

cjbell: So this is where I really think we eventually need to reconsider our guides into something anchored around Build, Test, Release, Observe, Improve, where this would fall under Release. That however, is something for the future.

Right now though, I think this is better placed under “Developer tools” as its own page, maybe with it renamed as “Integrating into CI/CD”

PR #396, opened 2024-02-06
fix: clarify push notification status updates

cellomatt:
This PR updates the delivery status documentation to clarify that you cannot update a push notification's delivery status, but you _can_ update the engagement status manually by using the `knock_message_id`.


rachael-t (approved):
Thanks for getting this in so quickly! 🎉

PR #397, opened 2024-02-06
chore: update documentation for commits and promotion 

cellomatt:
This PR updates the Environments concept page in our docs and moves the concepts of Commits and Promotion to a separate page. The changes include:

- Rearrange the order of some content to make it clearer
- Add additional search tags
- Add a new section on commit diffs (including a gif asset)
- Add explicit instructions on how to take certain actions in your dashboard, and mention the ability to perform these same actions via the MAPI or CLI
- Add Commits to sidebar nav
- Add Commits to the Concepts overview page, re-order the concepts on the overview page to match the nav ordering, and add a section for Recipients, which was missing
- Update the Workflows concept page's backlinks to separate version control and environments

### DIFF
--- a/content/concepts/commits.mdx
+++ b/content/concepts/commits.mdx
@@ -53,9 +53,9 @@ Changes must be made in the development environment, then staged and tested befo
       disable a workflow in any environment without needing to go through
       environment promotion. This is your kill switch for a given workflow
       should you need it; this status is displayed on each workflow in your
-      dashboard's "Workflows" section, and can
-      be set by clicking on the workflow and using the "Status" selector. It is
-      environment-specific and will only be applied to the current environment.
+      dashboard's "Workflows" section, and can be set by clicking on the
+      workflow and using the "Status" selector. It is environment-specific and
+      will only be applied to the current environment.
     </>
   }
 />
--- a/content/concepts/commits.mdx
+++ b/content/concepts/commits.mdx
@@ -1,0 +1,70 @@
+---
+title: Commits
+description: Learn about how Knock's commit and promotion model works.
+tags:
+  [
+    "branches",
+    "env",
+    "version control",
+    "versions",
+    "commit",
+    "promote",
+    "promotion",
+    "staging",
+    "active",
+    "inactive",
+    "diffs",
+  ]
+section: Concepts
+---
+
+To version the changes you make in your [environments](/concepts/environments), Knock uses a commit model. When you make a change to a workflow or a layout in the Knock dashboard, you'll need to commit it to your development environment before those changes will appear in workflows triggered via the API.
+
+After you modify a resource, you'll see a "Save" button that allows you to store those changes. When you're ready to permanently store your updates with version control, they should be committed with the "Commit to development" button that will come into focus after changes have been saved.
+
+**A few things to note:**
+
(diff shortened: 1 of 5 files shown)


cjbell (commented):
All of these changes in here look really great and I think really help explain the concept, but this makes me wonder: should we have commits as it's own page, or maybe we should call this page "Environments & commits". WDYT?

content/concepts/environments.mdx
```diff
@@ -23,12 +30,20 @@ The API key you use determines the environment into which you'll be sending data
...
-A couple other things to note:
+**A few things to note:**
 
 - Channel configurations, branding, and variables do not need to be committed, as they live at the account-level. This means that if you make a change to a channel configuration, it will update immediately on notifications sent in that environment.
-- Any changes you have not yet committed to your current environment **will** apply when you're using the test runner. This way you can test your latest changes before you commit them to your environment.
+- Any changes you have saved but not yet committed **will** apply when you're using the test runner. This way you can test your latest changes before you commit them to your development environment.
+- You can work with Knock resources outside of your dashboard if you prefer. We offer both a [Management API](/developer-tools/management-api) and a [command line interface](/developer-tools/knock-cli) for interacting with Knock resources programmatically.
+- The commit model applies to all methods of interacting with Knock resources, whether directly in the dashboard or with the Management API and CLI.
+
+### Commit Diffs
```
samseely, line 42:
```suggestion
### Commit diffs
```

Nit: this should be sentence case not title case

samseely:
> All of these changes in here look really great and I think really help explain the concept, but this makes me wonder: should we have commits as it's own page, or maybe we should call this page "Environments & commits". WDYT?

Yep agreed with this note from @cjbell . I'd lean towards "Commits" as its own page to keep as a clean separate concept that users can easily find. That page could also do a deeper dive walkthrough of the version control model in Knock. We'd probably move promotion/rollback into that (which makes sense give revert commit is coming out soon). When we get into branching that would be a good place for that information too.

cellomatt:
> > All of these changes in here look really great and I think really help explain the concept, but this makes me wonder: should we have commits as it's own page, or maybe we should call this page "Environments & commits". WDYT?
> 
> Yep agreed with this note from @cjbell . I'd lean towards "Commits" as its own page to keep as a clean separate concept that users can easily find. That page could also do a deeper dive walkthrough of the version control model in Knock. We'd probably move promotion/rollback into that (which makes sense give revert commit is coming out soon). When we get into branching that would be a good place for that information too.

Ha, I was having this same conversation with @rachael-t! Sounds good, will make that change 😄 

image

content/concepts/commits.mdx
```diff
@@ -0,0 +1,70 @@
+---
+title: Commits
```
cjbell, line 2: Now that this is a separate page, we also need to add to `data/sidebar.ts` under "Concepts" so it appears as a nav item in the sidebar

cjbell (approved):
one last comment but looking great

content/concepts/commits.mdx
cellomatt: So weird, I had this in my local version and somehow it didn't get tracked 🤔 anyway, added, thanks!

PR #399, opened 2024-02-07
chore: add install commands for currently-linked SDKs

cellomatt:
Per [this brief](https://www.notion.so/knockapp/Docs-Getting-Started-guides-30bdb89d33234613bad1bc31b4fca9f3), this PR adds install instructions for each of the SDKs that are currently linked from the docs site. We currently only have instructions for Node.

**Notes:**
- Some of these instructions (i.e. Java, Elixir, etc) require specific versions, which could create maintainability issues. Are we ok with this?
- None of our mobile SDKs are included at this time (unless you count the `build.gradle` instructions under `Java`)


data/code/sdks/install.ts
```diff
@@ -0,0 +1,54 @@
...
+  `,
+  go: `
+    go get github.com/knocklabs/knock-go/knock
+  `,
+  php: `
+    composer require knocklabs/knock-php php-http/guzzle7-adapter
+  `,
+  java: `
+  // To build.gradle:
+  dependencies {
```
thomaswhyyou, line 19: I think there's an extra indentation here after comment lines?

data/code/sdks/install.ts
```diff
@@ -0,0 +1,54 @@
...
+        <groupId>app.knock.api</groupId>
+        <artifactId>knock-client</artifactId>
+        <version>0.2.10</version>
+    </dependency>
+    <!-- more dependencies here -->
+  </dependencies>
+  `,
+  csharp: `
+  // Via the NuGet Package Manager
+  nuget install Knock.net
```
thomaswhyyou, line 36: Same here.

data/code/sdks/install.ts
```diff
@@ -0,0 +1,54 @@
...
+  dotnet add package Knock.net
+
+// Via Visual Studio IDE
+  Install-Package Knock.net
+  `,
+  elixir: `
+# Add to mix.exs
+def deps do
+  [
+    {:knock, "~> 0.4"}
```
thomaswhyyou, line 48: > Some of these instructions (i.e. Java, Elixir, etc) require specific versions, which could create maintainability issues. Are we ok with this?

Hm, not sure of a good automatic way to keep the versions up to date. We could add a comment in the code example about checking the latest version, though that's not awesome. 

> None of our mobile SDKs are included at this time (unless you count the build.gradle instructions under Java)

Probably need to check with @matgar26 on this. Regardless we can probably follow up in a separate PR to document mobile SDKs?

cellomatt: > Hm, not sure of a good automatic way to keep the versions up to date. We could add a comment in the code example about checking the latest version, though that's not awesome.
Yeah, not awesome. I suppose if we have some sort of runbook for releasing new SDK versions, updating this could get added in as a to-do?

> Probably need to check with @matgar26 on this. Regardless we can probably follow up in a separate PR to document mobile SDKs?
Agreed. I think it's probably best to match these examples to the SDKs that are available in the main dropdown menu.

cjbell (commented):
this looks awesome, but can we nix anything that has a static version and just replace it with a `VERSION` callout to keep this evergreen? I could see us never remembering to update the versions otherwise

cjbell (approved):

PR #402, opened 2024-02-09
chore: clarify batch functionality

cellomatt:
This PR attempts to clarify the functionality of batch functions, particularly sliding windows, and call out some of the caveats with batches that we either do not currently point out or reference only in passing. 

It also links adds a link back to these clarifications from the workflow cancellation documentation, and rewords one of the "gotchas" on that page to attempt to make it easier to understand.


content/designing-workflows/batch-function.mdx
```diff
@@ -95,21 +95,51 @@ When the key specified is missing, or resolves to an invalid value, a correspond
 
 ### Using a sliding batch window
 
-By default, all batch windows are fixed, where the closing window of the batch is determined by the first trigger that starts the batch. In some situations, you may wish to "extend" the batch window when a new trigger is received to recompute the closing time of the batch. This option is supported in the batch step as a "sliding window". Let's walk through an example:
+By default, all batch windows are fixed, where the closing of the batch window is determined by the first trigger that starts the batch. In some situations, you may wish to "extend" the batch window when a new trigger is received to recompute the closing time of the batch. This option is supported in the batch step as a "sliding window".
+
+When a sliding window is enabled on a batch function, subsequent workflow triggers that are detected by the already-open batch window will add the configured default window duration onto the already-open batch window. Let's walk through an example:
+
+**_Initial batch window: 1 minute_**
```
samseely, line 102: This confused me at first as currently formatted, just as we're using bold text to separate out other sections. 

Let's insert at top of list but put in brackets to make it clear it's part of the batch logic but is starting state. could use a 🎛️ emoji to communicate it's a config on the batch step.

content/designing-workflows/batch-function.mdx
```diff
@@ -180,6 +210,27 @@ On the Enterprise plan, you can configure the maximum number of `activities` and
 
 If you want to remove an item from a batch (example: a user deletes a comment), you can use our [workflow cancellation API](/send-notifications/canceling-workflows) to cancel a batched item, thereby removing it from the batch.
 
+<Callout
+  emoji="⛔"
+  text={
+    <>
+      <span className="font-bold">Important:</span> Once a batch window has been
+      opened, it will remain open until its full duration has elapsed. This is
+      because closing a batch window upon workflow cancellation would remove the
```
samseely, line 219: Let's articulate this more as how cancels interact with workflow runs within a batch, and less as that it would remove behavior if we had cancels close the batch window. 

Something like?
"Once a batch window is opened, it will remain open until its full duration has elapsed. Any workflow cancellations will remove the specific individual workflow run they reference from the batch.

Because of this behavior..."

samseely (approved):
A couple small notes but otherwise good to go!

PR #406, opened 2024-02-20
chore: add token expiring callback docs

cjbell:
### Description

- Add new docs for `onUserTokenExpiring`
- Add new section to in-app UI security on token refreshing

### DIFF
--- a/content/in-app-ui/security-and-authentication.mdx
+++ b/content/in-app-ui/security-and-authentication.mdx
@@ -172,7 +172,7 @@ knockClient.authenticate(currentUser.id, currentUser.knockToken);
 
 Generally, it's advisable to set the token expiration to be equal to your session token expiration. That way you can regenerate both of the tokens together from within your application. There may, however, be cases where this is not possible, in which case it's best practice to opt for a relatively short-lived expiration time for your user tokens and refresh them from your backend before the expiration window.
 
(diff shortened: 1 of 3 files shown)


connorlindsey (approved):

content/in-app-ui/security-and-authentication.mdx
```diff
@@ -156,21 +154,28 @@ import Knock from "@knocklabs/client";
...
   userId={currentUser.id}
   userToken={currentUser.knockUserToken}
 >
 ```
 
+## Handling token expiration
+
+Generally, it's advisable to set the token expiration to be equal to your session token expiration. That way you can regenerate both of the tokens together from within your application. There may, however, be cases where this is not possible, in which case it's best practice to opt for a relatively short-lived expiration time for your user tokens and refresh them from your backend before the expiration window.
+
+For convenience, the Knock JavaScript client and React package expose an `onUserTokenExpiring` callback method, that will be invoked before the user token expires. This provides an easy hook for your application to refresh the user token from your backend.
```
cellomatt, line 175:
```suggestion
For convenience, the Knock JavaScript client and React package expose an `onUserTokenExpiring` callback method which will be invoked before the user token expires. This provides an easy hook for your application to refresh the user token from your backend.
```

cellomatt (approved):

PR #412, opened 2024-02-27
fix(kno-5355): fix reference layout

meryldakin:
### Description
Fixes layout + adds missing type

### Tasks
[KNO-5355](https://linear.app/knock/issue/KNO-5355/fix-react-api-reference-layout)


JEverhart383 (approved):
LGTM

PR #414, opened 2024-02-29
Remove data prefix callout from segment doc

samseely:
Removing callout that you need to append "data." in place of "properties." to access variables that live within "properties" object of a segment track event. 

I just tested this and if you use `properties.` instead of `data.` it still works as expected, so I think we should remove this callout that you need to prefix properties with `data.` to use them in event action mappings. 

Here's a Loom of me testing this if you want to confirm: https://www.loom.com/share/e7dab22258f44ea996d2bad4089dc231

### DIFF
--- a/content/integrations/sources/segment.mdx
+++ b/content/integrations/sources/segment.mdx
@@ -59,7 +59,7 @@ From the source environment configuration page click the "View in environment" b
 
 ## Connecting a track event to a workflow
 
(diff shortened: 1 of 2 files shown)


samseely:
I've updated this callout a bit as even though `properties` works in event action mappings, think it's important to call out that within a template you access event properties directly. Might be worth a separate section in this doc on using track event props within a template.

cjbell (approved):

PR #419, opened 2024-03-06
chore: outbound webhooks clarifications

cellomatt:
WIP

### Description

** Describe what, why and how of the changes clearly and concisely. Add any additional useful context or info, as necessary. **

### Todos

** List any todo items necessary before merging, if any. Delete if none. **

- [ ] Sample todo item 1
- [ ] Sample todo item 2

### Tasks

** Link to task(s) which this PR corresponds to. Example: [KNO-54](https://linear.app/knock/issue/KNO-54/create-a-pr-template) **

### Screenshots

** Attach any screenshots or recordings to visually illustrate the changes, as necessary. Delete if not relevant. **

### DIFF
--- a/components/Attributes.tsx
+++ b/components/Attributes.tsx
@@ -4,12 +4,22 @@ const Attributes = ({ children }) => (
   </div>
 );
 
-const Attribute = ({ name, type, description }) => (
+const Attribute = ({ name, type, description, typeSlug }) => (
   <div className="attribute border-b dark:border-b-gray-800 py-2">
     <span>
       <span className="font-mono text-xs">{name}</span>
-      <span className="font-semibold text-gray-500 dark:text-gray-300 text-xs ml-2 py-0.5 px-1 border border-gray-100 bg-gray-50 dark:bg-transparent dark:border-transparent">
-        {type}
+      <span className="font-semibold text-gray-500 dark:text-gray-300 text-xs ml-2 py-0.5 px-1 border border-gray-100 bg-gray-50 dark:bg-gray-800 dark:border-gray-700">
+        {/* Pass an optional `typeSlug` to link the `type` to its definition in the docs*/}
+        {!!typeSlug ? 
+          <a 
+            href={typeSlug} 
+            className="not-prose transition-colors hover:text-gray-900 dark:hover:text-gray-100 text-brand hover:!text-brand-dark"
+            >
+            {type}
+          </a>
+        : 
+          type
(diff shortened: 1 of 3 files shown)


components/Attributes.tsx
```diff
@@ -4,12 +4,22 @@ const Attributes = ({ children }) => (
...
     <span>
       <span className="font-mono text-xs">{name}</span>
-      <span className="font-semibold text-gray-500 dark:text-gray-300 text-xs ml-2 py-0.5 px-1 border border-gray-100 bg-gray-50 dark:bg-transparent dark:border-transparent">
-        {type}
+      <span className="font-semibold text-gray-500 dark:text-gray-300 text-xs ml-2 py-0.5 px-1 border border-gray-100 bg-gray-50 dark:bg-gray-800 dark:border-gray-700">
+        {/* Pass an optional `typeSlug` to link the `type` to its definition in the docs*/}
+        {!!typeSlug ? (
+          <a
+            href={typeSlug}
+            className="not-prose transition-colors hover:text-gray-900 dark:hover:text-gray-100 text-brand hover:!text-brand-dark"
```
rachael-t, line 16: I agree that the dark mode styling could use some improvement, but it can be tricky because of accessibility issues with contrast. For example, this change with the red text causes those elements to be flagged in lighthouse checks. Could be worth asking Krisna for advice!

cellomatt: I messed around with this a bit and Lighthouse seems happy with contrast if I just leave the background transparent but keep the border. Slightly less-obvious but maybe easier to read:

image

rachael-t:
Love seeing the links to reference definitions! 👏 

My only feedback is what we chatted about in Slack:
- The contrast in dark mode with the style changes isn't _great_ and maybe just having the border would suffice?
- I don't know if it's possible to nest the `EventData` attributes, but it would be nice to have some sort of visual indicator as it currently feels a bit off in the flow.

components/Attributes.tsx
cellomatt: @ksorathia this is the PR I was referring to in standup this morning. I have added some styles to this Type label in our `Attribute` component, to make it a bit more obvious in dark mode; current styling sets `bg-transparent` and `border-transparent` in dark mode for that label (see screenshots in PR description). 

Do you have any thoughts around whether we should change current styling, and if so what color properties we should use here? 🙏

cjbell (approved):

PR #430, opened 2024-03-13
chore: add inkeep chat widget

cjbell:
### Description

Adds inkeep as a UI chat widget

### Tasks

[KNO-5437](https://linear.app/knock/issue/KNO-5437/add-inkeep-ai-bot-to-docs)


samseely (approved):
Looks good. A couple vqa thoughts in this loom that would be good to do now before we ship just since we may not be back here for a while https://www.loom.com/share/d8365bc7da304a0c8db1b8172b3275b8

PR #431, opened 2024-03-13
fix: show codeblock titles

cjbell:
### Description

Fix code block titles by using new package that parses them as props


samseely (approved):
Yay! An open question for us outside of this PR (for docs 2.0 brief), should all code examples have a title? Like is that something we should try to enforce through checks in PR review or something? Food for thought. I tend to think these code examples look weird/missing something w/o the title but understand that there could be exceptions

PR #432, opened 2024-03-14
chore: update example apps page

JEverhart383:
### Description
This PR reworks the example apps page using the `SDKCard` component treatment to make it easier to browse our different example apps. There's still some work to do to update all of the READMEs, but I feel like this is a decent foundation based on some of the examples we looked at and will support our near-term goals.

Screenshot 2024-03-14 at 3 16 27 PM


rachael-t (approved):
Love this update 🤩 So much nicer to just link to the example apps and have the step-by-step guide in the individual readmes

cellomatt:
Two comments/questions:
- Would it be better if we used icons for the various languages/frameworks rather than the same box icon on each one? I see that the SDKCard component just uses the same 
- I feel like it would be better to open these in a new tab rather than a redirect but that's my personal preference. Again I know this comes down to component functionality so maybe it's not great feedback for this specific PR

cellomatt (approved):

JEverhart383:
> Would it be better if we used icons for the various languages/frameworks rather than the same box icon on each one? I see that the SDKCard component just uses the same

Yeah, on a longer timeline we should do this. It will make it much more scannable 

> I feel like it would be better to open these in a new tab rather than a redirect but that's my personal preference. Again I know this comes down to component functionality so maybe it's not great feedback for this specific PR

I could also get on-board for this, and could be a simple component prop `useBlankTarget` or something

cjbell (approved):

PR #435, opened 2024-03-15
update 'design a notification' section in quick start

natperpepaj:
### Description

Update to the `Design a notification` section of the `Quick start` guide. The option to configure a template exists within the channel step "Edit step" view and the create button for the template previously mentioned no longer exists.


content/getting-started/quick-start.mdx
```diff
@@ -36,8 +36,8 @@ A workflow encapsulates a notification in Knock. Each workflow takes a trigger c
...
-- **Design a notification.** When your channel is configured, head back to the workflow's channel step and click "create" to create its notification template. The template starts with default copy, so we'll just use that for now.
+- **Add a channel step.** To send a notification, a workflow needs at least one channel step. A Knock channel represents a notification delivery medium (such as email or mobile push). To add this step, we'll click “edit steps” to enter the workflow canvas. Here we can see a number of steps available for us to add to our workflow including functions (such as batch and delay) and channels. Choose the channel you'd like to use in your workflow and follow the dashboard to configure the channel step.
```
meryldakin, line 39: I know this was already here but I think this part of the sentence doesn't make much sense anymore: ` and follow the dashboard to configure the channel step.`

Maybe we say `Choose the channel you'd like to use in your workflow and use the sidebar to configure the channel step` or something like that?

meryldakin (commented):
ty nat! looks good just a q about the existing wording

content/getting-started/quick-start.mdx
natperpepaj: I think this suggestion makes a lot of sense. It wasn't intuitive when I was trying to add mention of this to the design a notification step but I think this is the perfect place! Will update

cellomatt (approved):
🔥

PR #438, opened 2024-03-20
chore: simplify quick start guide

cellomatt:
This is a first pass at updating the Quick start in our docs. It doesn't change the overall approach for now, but it removes a lot of extraneous information that is covered elsewhere in the docs and adds additional links to more information about specific concepts. I also took the liberty of rewording a few things that seemed overly complex or redundant.


samseely:
Rewording updates look great and I'm glad we're taking out the big note on managing users. Only note I'd add here, after the call where we inline identify those users, might be worth calling out that we're inline identifying them? Maybe extraneous and we don't need to mention it but came to mind as I read through. Whatever feels right to you though, will leave it in your hands!

samseely (approved):

PR #443, opened 2024-03-26
chore: update quick start with new components

cellomatt:
### Description
This PR updates the Quick start guide to use the new `Step` and `Accordion` components.

### DIFF
--- a/content/getting-started/quick-start.mdx
+++ b/content/getting-started/quick-start.mdx
@@ -34,19 +34,19 @@ In this guide, you'll integrate Knock with your backend web application and send
 
     Here's how to build your first workflow:
     <AccordionGroup>
-      <Accordion title="Create a workflow.">
+      <Accordion title="Create a workflow">
         Click the "+ Workflow" button in the top right corner of the Knock dashboard. Name it whatever you like.
       </Accordion>
-      <Accordion title="Add a channel step.">
+      <Accordion title="Add a channel step">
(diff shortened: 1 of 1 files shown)


rachael-t:
YES! This is SO much easier on the eyes 👏  
Love the use of the new `Step` component! I'm a little on the fence with the `Accordion` component in step 4. While this helps a lot with the initial digestion of the page (and then as a user I think, "Wow, this is so short and quick - I'll definitely dive into this!"), I just don't know if the accordion totally feels right. I personally lean more towards moving forward with it, but can't quite put my finger on why it feels awkward. Curious about what others think!

content/getting-started/quick-start.mdx
```diff
@@ -7,67 +7,79 @@ section: Getting started
...
+
+  </Step>
+  <Step title="Create a workflow in Knock">
+    Next we'll design our first workflow in Knock via the dashboard.
+
+    A workflow encapsulates a notification in Knock. Each workflow takes a trigger call via the Knock API, runs the data you provide through a set of logic you configure, and outputs the actual messages that will be sent to your end users. All channel routing and message design takes place within the workflow.
+
+    Here's how to build your first workflow:
+    <AccordionGroup>
+      <Accordion title="Create a workflow.">
```
cjbell, line 37: Nit: I don't think these accordion titles should have periods

cjbell (approved):
one nit but I love this and cannot wait to ship it

PR #444, opened 2024-03-27
fix: <aside> component height

cellomatt:
### Description

This PR adjusts the height calculation of the `<aside>` menu component so that it never overlaps the footer.

### Video explanation

https://www.loom.com/share/e371409be70c459f902a5b751ca9e5eb?sid=8f7a53b0-e4b6-4813-a3f5-4fa17d1d1e74


rachael-t (approved):
👏

PR #447, opened 2024-03-28
add a couple br to quickstart accordions

samseely:
added a couple breaks to copy in accordions to give it a bit more whitespace


cellomatt (approved):
🔥

PR #452, opened 2024-04-01
chore(kno-5632): update api docs with recipients and objects filtering

natperpepaj:
### Description

Update api docs to represent `recipients` and `objects` filtering as a map of maps. For recipients, both list and map structures are accepted

### Tasks

Ticker [KNO-5632](https://linear.app/knock/issue/KNO-5632/[docs]-update-objects-and-recipients-support-for)

### DIFF
--- a/content/reference.mdx
+++ b/content/reference.mdx
@@ -4554,9 +4554,14 @@ Will return the underlying recipient attached. Note: the object must exist.
     description='When set to "recipient" will return all of the active subscriptions for which the specified object is a recipient (as a subscriber to other objects), rather than the subscribers of the specified object.'
   />
   <Attribute
+    name="objects"
+    type="dictionary<string, ObjectRef>"
+    description='A collection of ObjectRefs, containing the object collection and id, to filter subscriptions. Filtering by objects will only be enabled when mode is set to "recipient".'
+  />
+  <Attribute
     name="recipients"
-    type="string[]"
-    description="A list of recipient identifiers to filter subscribers of the object. Filtering by recipients will only be enabled when mode is not set."
+    type="string[] | dictionary<string, RecipientRef>"
+    description="A collection of recipient identifiers (user_ids and/or ObjectRefs) to filter subscribers of the object. Filtering by recipients will only be enabled when mode is not set."
   />
 </Attributes>
 
@@ -4638,6 +4643,11 @@ Returns a paginated list of subscriptions for a single user, in descending order
     name="before"
     type="string"
     description="The cursor to retrieve items before (hint: use the `__cursor` field)"
+  />
+  <Attribute
+    name="objects"
+    type="dictionary<string, ObjectRef>"
+    description="A collection of ObjectRefs, containing the object collection and id, to filter subscriptions."
   />
 </Attributes>
 
@@ -5089,8 +5099,8 @@ Returns a paginated list of schedules in descending order.
   />
   <Attribute
     name="recipients"
-    type="string[]"
-    description="A list of recipient identifiers to filter schedules for."
+    type="string[] | dictionary<string, RecipientRef>"
+    description="A collection of recipient identifiers (user_ids and/or ObjectRefs) to filter schedules for."
   />
   <Attribute
     name="tenant"


content/reference.mdx
```diff
@@ -4553,10 +4553,15 @@ Will return the underlying recipient attached. Note: the object must exist.
     type="string"
     description='When set to "recipient" will return all of the active subscriptions for which the specified object is a recipient (as a subscriber to other objects), rather than the subscribers of the specified object.'
   />
+  <Attribute
+    name="objects"
+    type="map<string, user_id | object>"
```
cjbell, line 4558: I think we usually use `dictionary` here and I think we can make this `dictionary<string, RecipientRef>`

content/reference.mdx
```diff
@@ -4553,10 +4553,15 @@ Will return the underlying recipient attached. Note: the object must exist.
...
+    name="objects"
+    type="dictionary<string, ObjectRef>"
+    description='A collection of objectRefs, containing the object collection and id, to filter subscriptions. Filtering by objects will only be enabled when mode is set to "recipient".'
+  />
   <Attribute
     name="recipients"
-    type="string[]"
-    description="A list of recipient identifiers to filter subscribers of the object. Filtering by recipients will only be enabled when mode is not set."
+    type="string[] | dictionary<string, RecipientRef>"
+    description="A collection of recipient identifiers, userIds and/or objectRefs, to filter subscribers of the object. Filtering by recipients will only be enabled when mode is not set."
```
cellomatt, line 4564: Not sure if the casing/capitalization difference is intentional here?

```suggestion
    description="A collection of recipient identifiers (user_ids and/or ObjectRefs) to filter subscribers of the object. Filtering by recipients will only be enabled when mode is not set."
```

content/reference.mdx
```diff
@@ -4553,10 +4553,15 @@ Will return the underlying recipient attached. Note: the object must exist.
     type="string"
     description='When set to "recipient" will return all of the active subscriptions for which the specified object is a recipient (as a subscriber to other objects), rather than the subscribers of the specified object.'
   />
+  <Attribute
+    name="objects"
+    type="dictionary<string, ObjectRef>"
+    description='A collection of objectRefs, containing the object collection and id, to filter subscriptions. Filtering by objects will only be enabled when mode is set to "recipient".'
```
cellomatt, line 4559:
```suggestion
    description='A collection of ObjectRefs, containing the object collection and id, to filter subscriptions. Filtering by objects will only be enabled when mode is set to "recipient".'
```

content/reference.mdx
```diff
@@ -4639,6 +4644,11 @@ Returns a paginated list of subscriptions for a single user, in descending order
     type="string"
     description="The cursor to retrieve items before (hint: use the `__cursor` field)"
   />
+  <Attribute
+    name="objects"
+    type="dictionary<string, ObjectRef>"
+    description="A collection of objectRefs, containing the object collection and id, to filter subscriptions."
```
cellomatt, line 4650: Same q as above re: `ObjectRefs`

content/reference.mdx
```diff
@@ -5089,8 +5099,8 @@ Returns a paginated list of schedules in descending order.
   />
   <Attribute
     name="recipients"
-    type="string[]"
-    description="A list of recipient identifiers to filter schedules for."
+    type="string[] | dictionary<string, RecipientRef>"
+    description="A collection of recipient identifiers, userIds and/or objectRefs, to filter schedules for."
```
cellomatt, line 5103: Same as above:

```suggestion
    description="A collection of recipient identifiers (user_ids and/or ObjectRefs) to filter schedules for."
```

cellomatt (approved):
I think this is good 👌 Maybe we can also include updated code examples (once SDKs are all shipped) that show usage of the filters?

content/reference.mdx
natperpepaj: no intention there just a miss on my part - as for the user_ids was just keeping the camelcase consistency but we reference it elsewhere as snakecased so will commit your suggested changes. thanks for checking it out!

natperpepaj:
> I think this is good 👌 Maybe we can also include updated code examples (once SDKs are all shipped) that show usage of the filters?

definitely!

PR #455, opened 2024-04-01
fix: typo on setting preferences

rachael-t:
Small typo fix 😄


cellomatt (approved):
lol I kind of like **Example an example** 😆 

🚀

PR #457, opened 2024-04-02
chore: clarify inline preference set

cellomatt:
### Description

This PR updates our documentation for setting preferences inline. In particular, it makes it a bit clearer that you must explicitly provide the `default` key when setting preferences this way, unlike the other ways of setting preferences (when using one of our SDKs), which will default to `default` when no preference set ID is provided. I also moved the callout block above the code example, as I think it was likely burying the lede a bit before.

### DIFF
--- a/content/managing-recipients/setting-preferences.mdx
+++ b/content/managing-recipients/setting-preferences.mdx
@@ -109,28 +109,29 @@ In the case in which the per-tenant preference set doesn't exist, the [`default`
 
 ### Setting preferences inline during a workflow trigger
 
-Preferences can also be set for one or more recipients during a workflow trigger using [inline identify](/send-notifications/triggering-workflows#identifying-recipients-inline).
+<Callout
+  emoji="🚨"
+  text={
+    <>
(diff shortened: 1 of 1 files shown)


content/managing-recipients/setting-preferences.mdx
```diff
@@ -109,29 +109,29 @@ In the case in which the per-tenant preference set doesn't exist, the [`default`
...
 <Callout
   emoji="🚨"
   text={
     <>
-      <strong>Please note</strong> setting preferences inline will always upsert
-      the preference set, meaning that changes will be overidden. Generally, we
-      recommend against using inline preferences to ensure the integrity of your
-      data.
+      <strong>Please note:</strong> setting preferences inline will always
+      upsert the preference set, meaning that changes will overide any extant
```
connorlindsey, line 117: Nit: Could we find a simpler alternative to extant here? I think something like 'existing' would work well and be more easily understood.

connorlindsey (approved):
Looks good to me! Just one wording comment.

cjbell (approved):
great change

PR #458, opened 2024-04-02
chore: add new components to slackkit setup

cellomatt:
### Description

This PR adds the new Step and Accordion components to the SlackKit setup page to organize things better. It also makes a few copy edits for clarity and grammar, and removes `code` styling from non-code references.

https://docs-o4ex5sir5-knocklabs.vercel.app/integrations/chat/slack-kit/setup

### DIFF
--- a/content/integrations/chat/slack-kit/setup.mdx
+++ b/content/integrations/chat/slack-kit/setup.mdx
@@ -42,7 +42,7 @@ In this section we'll complete the steps for setting up your Knock account and S
         You can read more about scopes in general in [building a slack app](/integrations/chat/slack/building-a-slack-app).
       </Accordion>
       <Accordion title="Add redirect URL">
-        Under `OAuth & Permissions` in the `Features` sidebar, you need to add a redirect URL so that Knock can receive the Slack OAuth handshake on your behalf.
+        Under **OAuth & Permissions** in the **Features** sidebar, you need to add a redirect URL so that Knock can receive the Slack OAuth handshake on your behalf.
 
         Copy and paste this redirect URL:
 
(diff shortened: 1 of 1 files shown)


content/integrations/chat/slack-kit/setup.mdx
```diff
@@ -8,65 +8,71 @@ layout: integrations
...
+      }
+    />
+
+    The following steps will be completed in the **OAuth & Permissions** sidebar of your Slack app's [management page](https://api.slack.com/apps):
+
+    <AccordionGroup>
+      <Accordion title="Add a scope">
+        Although SlackKit manages your scopes, you have to add one in your Slack application's settings to start with so that it exposes the form for you to add a redirect URL.
+
+        Add `channels:read` to the scopes section. For the rest, you won't be managing SlackKit's scopes directly since the components will request the scopes they need. Here is the list of scopes and they reasons they're required:
```
rachael-t, line 35:
```suggestion
        Add `channels:read` to the scopes section. For the rest, you won't be managing SlackKit's scopes directly since the components will request the scopes they need. Here is the list of scopes and the reasons they're required:
```

rachael-t (commented):
I freaking love these components 🤩 this looks great!

Something I think we should consider reformatting on this page (and possibly elsewhere in the docs) is limiting inline code formatting only when referencing code or identifiers, and not necessarily for emphasis. For example, we use it here to help emphasize/make buttons/locations in the UI clear to the user as they are following along, but maybe just reformatting from `Features` to "Features" would work. 

Also, I added a comment for super small typo fix 🤓

cellomatt:
> Something I think we should consider reformatting on this page (and possibly elsewhere in the docs) is limiting inline code formatting only when referencing code, identifiers, etc., and not necessarily for emphasis.

@rachael-t I 100% agree, I'll go back and make these updates. I didn't actually update the original content (just moved into the components) but I think that's a great callout.

rachael-t (approved):
Thanks for making those changes! It looks great! 🙌

PR #463, opened 2024-04-10
chore: add support for dynamic language list on multilang code block

cellomatt:
### Description

In preparation for adding `cURL` and mobile SDK examples to some of our multi-language code block examples (in the API ref), this PR does several things:

- register `cURL`, Kotlin, and Swift as `SupportedLanguages`
- refactor `MultiLanguageCodeBlock` to base the list of languages in the switcher on the examples in the given code `snippet`, ordered by a `DEFAULT_ORDER`
- add additional logic to the rendered `code` so that if a given code block does not include an example for the language that is currently stored in `localstorage`, we will still render the correct code to match the language that is displayed on the switcher (which will always be the first language in the list).
For example: if you switch a block to `cURL` and `"curl"` is subsequently set in local storage, with the current component any block which doesn't include example `curl` code will display as empty even though the switcher will show the first language in its list.
- add a usage example and explanation to the `/playground` page

### Questions

1. What should the `DEFAULT_ORDER` be? Do we have any sense of the popularity of certain SDKs after Node? For now I essentially kept the order the same as before, adding `cURL` and `shell` at the beginning (with the assumption that these should be first when present) and everything else at the end of the list.

### DIFF
--- a/components/MultiLangCodeBlock.tsx
+++ b/components/MultiLangCodeBlock.tsx
@@ -3,7 +3,7 @@
 */
 
 import { useEventEmitter } from "@byteclaw/use-event-emitter";
-import { useEffect, useMemo, useState } from "react";
+import { useEffect, useMemo } from "react";
 import { useIsMounted } from "../hooks/useIsMounted";
 import useLocalStorage from "../hooks/useLocalStorage";
 import { CodeBlock, SupportedLanguage } from "./CodeBlock";
--- a/components/MultiLangCodeBlock.tsx
+++ b/components/MultiLangCodeBlock.tsx
@@ -11,7 +11,8 @@ import { CodeBlock, SupportedLanguage } from "./CodeBlock";
 type Props = {
   title: string;
   snippet: string;
-  languages?: string[];
+  languages?: SupportedLanguage[]; //optionally define the list of languages included in the language switcher
+  includeDefault?: boolean; //whether or not to include DEFAULT_LANGUAGES in the switcher; defaults to true when not provided
 };
 
 const LOCAL_STORAGE_KEY = "@knocklabs/example-lang";
@@ -111,6 +112,8 @@ const snippets = {
     require("../data/code/workflows/trigger-with-user-channel-data").default,
   "workflows.trigger-with-user-preferences":
     require("../data/code/workflows/trigger-with-user-preferences").default,
+    "workflows.playground":
+    require("../data/code/workflows/playground").default,
 };
 /* eslint-enable */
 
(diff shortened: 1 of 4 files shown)


components/MultiLangCodeBlock.tsx
```diff
@@ -11,7 +11,8 @@ import { CodeBlock, SupportedLanguage } from "./CodeBlock";
 type Props = {
   title: string;
   snippet: string;
-  languages?: string[];
+  languages?: SupportedLanguage[]; //optionally define the list of languages included in the language switcher
+  includeDefault?: boolean; //whether or not to include DEFAULT_LANGUAGES in the switcher; defaults to true when not provided
```
cjbell, line 15: Do we need both of these properties? Could we just support `languages` and then set the default to be `DEFAULT_LANGUAGES` instead?

components/MultiLangCodeBlock.tsx
```diff
@@ -125,14 +127,27 @@ const DEFAULT_LANGUAGES: SupportedLanguage[] = [
...
     "node",
   );
 
+  // Create the combined list of explicitly provided languages plus the default languages list, and remove any duplicates
+  if (languages && includeDefault) {
+    const combinedLangs = DEFAULT_LANGUAGES.concat(languages);
+    languages = combinedLangs.filter(
+      (item, pos) => combinedLangs.indexOf(item) === pos,
+    );
+  }
```
cjbell, line 149: I think with my comment above that will simplify the code here

components/MultiLangCodeBlock.tsx
cellomatt: I considered that, but the way I was thinking about it is: 
- you might sometimes want to define a list of just a few `languages` (say, only Swift and Kotlin for mobile SDK examples). This works great with only the `languages` prop
- You might separately want to do something like add a `cURL` example to the API ref for an endpoint, but not add `cURL` to the `DEFAULT_LANGUAGES` list because it might not be relevant to every code example that is using the defaults

Without `includeDefault`, for the second scenario you would have to refactor every existing example that is using the default languages and explicitly list all of the `languages`, which didn't seem like a great solution either 🤷

cellomatt: Maybe the code comment could have been clearer, but `includeDefault` is meant to be used when the defaults are additive to whatever you're explicitly defining in `languages`. The `CodeBlock` component will still use `DEFAULT_LANGUAGES` when no `languages` are provided.

cjbell: Hmm thinking about this more though for the two scenarios you laid out, we could actually make the code block _sample aware_ so that it knows that `workflows.playground` is the snippet and that snippet has N supported languages, so those are the languages listed. That would work for both scenario 1 + 2 here and means we never show an empty code example when the snippet doesn't have all languages as well? WDYT?

cellomatt: Seems reasonable to me! Only caveat there would be that the order of the list would depend on whoever creates the snippet following the same pattern. That's also true of the `languages` array but if `DEFAULT_LANGUAGES` applies to _most_ scenarios, at least those ones would be consistent. I suppose we could alphabetize the languages we get from the snippet?

cjbell: That's true. We could have another order list and order the snippets lang list consistently by that sorted list 💡

cjbell (approved):

PR #466, opened 2024-04-15
chore: add callout on finding channel id for Discord

cellomatt:
### Description

This PR adds a callout block to help a user understand what the Knock `channel_id` is and where to find it for a Discord channel (similar to the callout for setting channel data for Slack). This is in response to customer confusion on what this ID represents.

image

### DIFF
--- a/content/integrations/chat/discord.mdx
+++ b/content/integrations/chat/discord.mdx
@@ -56,6 +56,22 @@ Take the URL you copied from the Discord webhook and set channel data on the Obj
   title="Store Discord connection on object"
 />
 
+<br />
+
+<Callout
+  emoji="🚨"
+  text={
+    <>
+      <span className="font-bold">Potential confusion alert.</span> In the
+      example above, the KNOCK_DISCORD_CHANNEL_ID variable is the id of the
+      Knock channel you've created to represent your Discord integration within
+      the Knock dashboard. You can get it by going to{" "}
+      {"Integrations > Channels"} in the Knock dashboard and then copying the id
+      of your Discord channel.
+    </>
+  }
+/>
+
 You can navigate back to your Objects dashboard and verify that `project-1` has the appropiate data set.
 
 Now you're ready to send notifications to Discord via incoming webhook! Jump to the [Triggering the workflow](/integrations/chat/discord#triggering-the-workflow) section to do so.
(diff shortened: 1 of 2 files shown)


content/integrations/chat/discord.mdx
```diff
@@ -56,6 +56,22 @@ Take the URL you copied from the Discord webhook and set channel data on the Obj
...
 />
 
+<br />
+
+<Callout
+  emoji="🚨"
+  text={
+    <>
+      <span className="font-bold">Potential confusion alert.</span> In the
+      example above, the KNOCK_DISCORD_CHANNEL_ID variable is the id of the
```
JEverhart383, line 66:
```suggestion
      example above, the <code>KNOCK_DISCORD_CHANNEL_ID</code> variable is the id of the
```
Nit: If I reference something from a code sample inline, I try to format it as code, but YMMV.

content/integrations/chat/discord.mdx
```diff
@@ -56,6 +56,22 @@ Take the URL you copied from the Discord webhook and set channel data on the Obj
...
+      Knock channel you've created to represent your Discord integration within
+      the Knock dashboard. You can get it by going to{" "}
```
JEverhart383, line 68:
```suggestion
      the Knock dashboard. You can find it by going to{" "}
```

JEverhart383 (approved):
LGTM, just a couple of nits. I wonder how we could make this concept more clear 'globally' if that's possible

cellomatt:
> LGTM, just a couple of nits. I wonder how we could make this concept more clear 'globally' if that's possible

Yeah, same! I basically copy-pasted this from the Slack example -- I think we have it there (and I thought it would be good to add here) because when folks are only interested in a specific integration, sometimes they'll jump straight into these step-by-step guides for a quick implementation test before really reading about the rest of how Knock works. I'll think about if there are other ways/places we can call that out, but I agree that this should ideally be more obvious.

I do think one other possible point of confusion that is specific to chat apps is the whole "channel" thing, because we have the concept of a Knock channel but also the chat app needs to deliver a notification to a specific channel in the configured destination. If you're not super familiar with a channel in Knock, it gets confusing quickly

content/integrations/chat/discord.mdx
cellomatt: I like this, I think I'll go back and fix it in the Slack example too 🙏

PR #468, opened 2024-04-16
chore:reorder pref headings

JEverhart383:
### Description

quick reorder of headings based on recent questions about environment defaults


cellomatt (approved):
lgtm. I think we should also update line 220 to say `"Developers > Preferences"` rather than `"Developers > Defaults"` and merge this when the Switchboard update is released

https://github.com/knocklabs/docs/pull/468/files#diff-ddf9c79d876ceecc871df8785dca121c48bdbd6464a591869de9476c431216a1R220

PR #469, opened 2024-04-16
chore: clarify object vs. user recipients for slack

cellomatt:
### Description

This PR attempts to make it easier to find information on notifying Slack users via DM, as well as making a few copy edits and cosmetic improvements to the Slack integration docs. It also calls out the fact that SlackKit is designed for the channel selection/notification use case (see SlackKit overview page).


samseely (approved):
Great updates

rachael-t (approved):
Very helpful clarifications! 🙌

PR #471, opened 2024-04-18
chore: update segment source doc page

rachael-t:
### Description

I updated the Segment source page to use the new step and accordion components, made a few other formatting changes, and added an FAQ section.

Would love feedback on the two FAQ items. Specifically, I debated the following when thinking about adding this and where:
- The process is essentially the same, so should we combine them and adjust the language?
- Does it make sense to have them as FAQs, or should we instead add them as a section (possibly after "Committing the event action configuration")?

Preview: https://docs-git-rt-update-segment-source-events-knocklabs.vercel.app/integrations/sources/segment

### DIFF
--- a/content/integrations/sources/segment.mdx
+++ b/content/integrations/sources/segment.mdx
@@ -51,7 +51,7 @@ To start routing your Segment events to Knock, click the "Create Source" button 
   <Step title="Add mapping">
     At the top of the page, navigate to the **Mappings** tab. Click "New Mapping", and under Actions, click "Send".
 
-    You will then be directed to a form to configuring the webhook:
+    You will then be directed to a form to configure the webhook:
     <AccordionGroup>
       <Accordion title="Select events to map and send">
         If you'd like to send all events to Knock, you can add the condition **Event type is Track** and a condition for **Event type is Identify.** Make sure you change the top operator to **any** instead of all.
@@ -132,7 +132,7 @@ Event action configurations are stored in the commit history, just like workflow
 
 ## Disabling an action
 
-Actions are automatically enabled when you create them. To stop an event from triggering an action, go to the action page and toggle its status to **Inactive.** Keep in mind that this will disable that action for the current environment only. When you're ready to trigger the action again, you can set it back to **Active.**
+Actions are automatically enabled when you create them. To stop an event from triggering an action, go to the action page and toggle its status to **Inactive.** Keep in mind that this will only disable that action for the current environment. When you're ready to trigger the action again, you can set it back to **Active.**
 
 ## Enabling Identify events
 
@@ -223,21 +223,24 @@ You can learn more about inline identification in [our users concept guide](/sen
   ></iframe>
 </div>
 
+<br />
+
 ## Frequently asked questions
 
 <AccordionGroup>
-  <Accordion title="Can I have one event trigger multiple workflows?">
+  <Accordion title="Can I use a single Segment event to trigger multiple workflows?">
     Yes. Under the **Developers** > **Sources** section in your Knock dashboard,
     select an event from Segment to view a list of actions configured for that
     event. Then, in the upper right-hand corner, click the "Create action"
     button to select the additional workflow you want this event to trigger,
     then click "Create." You can then make any changes to the schema mapping
-    before saving and committing this new action to your development
+    before saving and committing this new action to your Development
     environment.
   </Accordion>
-  <Accordion title="Can I have multiple events trigger one workflow?">
-    Yes. To do so, you will follow the steps above to create an action for the
-    additional event and select the same workflow from the **Create workflow
-    trigger mapping** modal.
+  <Accordion title="Can I use multiple different Segment events to trigger the same workflow?">
+    Yes. To do so, go to the **Developers** > **Sources** section in your Knock
+    dashboard, select an event, and then click the "Create action" button in the
+    upper right-hand corner. You will choose the same workflow from the **Create
+    workflow trigger mapping** modal.
   </Accordion>
 </AccordionGroup>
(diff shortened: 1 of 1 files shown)


content/integrations/sources/segment.mdx
```diff
@@ -29,60 +29,82 @@ You can start routing your Segment events to Knock by creating a source of type
...
+      </Accordion>
+    </AccordionGroup>
+  </Step>
+  <Step title="Save changes">
+    Now you'll see a **Basic Settings** page. Check the "Enable Destination" button and click "Save changes."
+  </Step>
+  <Step title="Add mapping">
+    At the top of the page, navigate to the **Mappings** tab. Click "New Mapping", and under Actions, click "Send".
+
+    You will then be directed to a form to configuring the webhook:
```
cellomatt, line 54:
```suggestion
    You will then be directed to a form to configure the webhook:
```

content/integrations/sources/segment.mdx
```diff
@@ -110,13 +132,13 @@ Event action configurations are stored in the commit history, just like workflow
 
 ## Disabling an action
 
-Actions are automatically enabled when you create them. If you want to stop an event from triggering an action, you can go to the action page and toggle its status to "Inactive". Keep in mind that this will disable that action for the current environment only. When you're ready to trigger the action again, you can set it back to "Active".
+Actions are automatically enabled when you create them. To stop an event from triggering an action, go to the action page and toggle its status to **Inactive.** Keep in mind that this will disable that action for the current environment only. When you're ready to trigger the action again, you can set it back to **Active.**
```
cellomatt, line 135: Personal preference nit:
```suggestion
Actions are automatically enabled when you create them. To stop an event from triggering an action, go to the action page and toggle its status to **Inactive.** Keep in mind that this will only disable that action for the current environment. When you're ready to trigger the action again, you can set it back to **Active.**
```

content/integrations/sources/segment.mdx
```diff
@@ -200,3 +222,22 @@ You can learn more about inline identification in [our users concept guide](/sen
     }}
   ></iframe>
 </div>
+
+## Frequently asked questions
+
+<AccordionGroup>
+  <Accordion title="Can I have one event trigger multiple workflows?">
```
cellomatt, line 229: I feel like this could be worded slightly differently but I'm not sure exactly what about it bugs me; feel free to take it or leave it :) 
```suggestion
  <Accordion title="Can I use a single Segment event to trigger multiple workflows?">
```

content/integrations/sources/segment.mdx
```diff
@@ -200,3 +222,22 @@ You can learn more about inline identification in [our users concept guide](/sen
...
+    Yes. Under the **Developers** > **Sources** section in your Knock dashboard,
+    select an event from Segment to view a list of actions configured for that
+    event. Then, in the upper right-hand corner, click the "Create action"
+    button to select the additional workflow you want this event to trigger,
+    then click "Create." You can then make any changes to the schema mapping
+    before saving and committing this new action to your development
```
cellomatt, line 235:
```suggestion
    before saving and committing this new action to your Development
```

content/integrations/sources/segment.mdx
```diff
@@ -200,3 +222,22 @@ You can learn more about inline identification in [our users concept guide](/sen
...
+    environment.
+  </Accordion>
+  <Accordion title="Can I have multiple events trigger one workflow?">
```
cellomatt, line 238: Similar to above, take it or leave it:
```suggestion
  <Accordion title="Can I use multiple different Segment events to trigger the same workflow?">
```

content/integrations/sources/segment.mdx
```diff
@@ -200,3 +222,22 @@ You can learn more about inline identification in [our users concept guide](/sen
...
+    Yes. To do so, you will follow the steps above to create an action for the
```
cellomatt, line 239: Just calling out that if the above accordion is closed, "the steps above" might not be fully obvious

cellomatt (approved):
Yay this is great! Added a few nits but this will be a fab update

content/integrations/sources/segment.mdx
```diff
@@ -29,60 +29,82 @@ You can start routing your Segment events to Knock by creating a source of type
...
+  <Step title="Configure webhook">
+    Click the "Configure Webhooks (Actions)" button in the top right to enter the setup flow to configure your webhook:
+    <AccordionGroup>
+      <Accordion title="Select data source">
+        Check the data source you'd like to receive events from and click "Next."
+      </Accordion>
+      <Accordion title="Setup">
+        Give a name to the destination, like "Knock", and keep the "Fill in settings manually" option selected, then click "Create destination."
+      </Accordion>
+    </AccordionGroup>
```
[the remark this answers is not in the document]
rachael-t: @cellomatt I've also been on the fence about the best practice for accordions. They are really nice at organizing the page, but I do agree that sometimes it doesn't make sense.

I just pushed up an update without _without_ the accordions, which you can view here: https://docs-ip9rkfnua-knocklabs.vercel.app/integrations/sources/segment
And then this is the old version _with_ the accordions: https://docs-djfqltido-knocklabs.vercel.app/integrations/sources/segment

I lean towards removing them, but let me know your thoughts!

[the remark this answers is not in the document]
cellomatt: Oh thanks, that's really helpful to visualize!

I think in this case I also like it better without. I'll keep this bullet-point alternative in mind, it could be the better approach in some situations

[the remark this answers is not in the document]
rachael-t: Cool, I'll go ahead and get this merged in then 🎉 

Yeah I think with time we will have a better feel for the best practices for the components and can feel confident about ironing out the style guide 💪

PR #473, opened 2024-04-19
chore: add support to open card links in new tab

cellomatt:
### Description

Adds the option to open a card component's link in a new tab.

https://www.loom.com/share/cc6ab01d1fd14c11ae90e32ba30c285e?sid=f40e8424-d220-44d1-99c2-c5fa21f831bf

### DIFF
--- a/components/Card.tsx
+++ b/components/Card.tsx
@@ -6,7 +6,7 @@ type Props = {
   title: string;
   linkUrl: string;
   footer?: ReactElement;
-  newTab?: boolean;
+  isExternal?: boolean;
 };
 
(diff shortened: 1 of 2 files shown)


components/Card.tsx
```diff
@@ -6,6 +6,7 @@ type Props = {
   title: string;
   linkUrl: string;
   footer?: ReactElement;
+  newTab?: boolean;
```
cjbell, line 9: One pattern that’s nice for this prop is calling it “external” or “isExternal” and then having that dictate the new window behavior

cjbell (approved):

PR #475, opened 2024-04-22
chore: preferences clarifications

cellomatt:
### Description

This PR attempts to make our Preferences concept page easier to understand by doing some copy editing, reorganizing, and adding some components to make things neater. 

One open question is whether (and how) we should split this page into smaller pages, or move some of this content elsewhere. We do have a separate guide on [setting preferences](https://docs.knock.app/managing-recipients/setting-preferences) that is also quite lengthy, so I'm not sure if moving things there is necessarily the right move.

I also considered adding additional accordions to some of the content on this page (for example, the types of preferences, or the individual sections under "Preference defaults"), but doing so would mean losing the side nav anchors.

https://docs-g6xo9pbov-knocklabs.vercel.app/concepts/preferences


JEverhart383:
I'll come back through with some additional suggestions, but just wanted to offer my 2 cents on your open question. I wonder if we could lift the code samples from '/managing-recipients/setting-preferences' to '/concepts/preferences' to better support some of those examples and collapse those pages. Then I think the 'client integration' JS code from '/managing-recipients/setting-preferences' would actually make more sense to live in '/in-app-ui/javascript/preferences'

content/concepts/preferences.mdx
```diff
@@ -5,33 +5,66 @@ tags: ["recipients", "conditions", "prefs"]
 section: Concepts
 ---
 
-A [Preference](/reference#preferences) determines whether a given recipient (a user or an object) will receive a notification for a given workflow, channel, or workflow-channel pair.
-
-In this guide we'll walk through how to set notification preferences for your recipients in Knock.
+A [Preference](/reference#preferences) determines whether a given recipient (a user or an object) will receive a notification for a given workflow, channel, or workflow-channel pair. In this guide we'll walk through how to set notification preferences for your recipients in Knock.
 
 We'll start with a quick overview of the preferences model and how it ties into the rest of Knock, then we'll walk through how to set notification preferences for workflows, categories, and channels, as well as combinations of all three.
```
JEverhart383, line 10:
```suggestion
We'll start with a quick overview of the preferences model and how it ties into the rest of Knock, then we'll walk through how to set notification preferences for workflows, categories, and channels, as well as combinations of all three.
```
I wonder if we really need this paragraph? This statement isn't inclusive of all the topics in the doc, but we have the jump menu that already gives the reader an outline of what we'll cover. Just trying to really question every word as we pare this down.

content/concepts/preferences.mdx
```diff
@@ -5,33 +5,66 @@ tags: ["recipients", "conditions", "prefs"]
...
+      for a given channel, workflow, or workflow category, Knock defaults them
+      to <code className="font-bold">true</code>.
+    </>
+  }
+/>
+
+### Where preferences can be set
 
-When the channel step executes in your workflow run, the recipient will only receive a notification if their preferences for the channel type, the workflow, and the workflow category all evaluate to `true`. To put this in inverse terms, if **ANY** preference related to a given channel step in a workflow evaluates to `false`, we will not send the recipient a notification. This makes sense, as receiving a notification you thought you opted out of is not a fun experience for recipients. 😁 (Note: if you do not set a preference for a given channel, workflow, or category, Knock defaults them to `true`.)
+Preferences can be set at three different levels in Knock:
```
JEverhart383, line 49: I like this explicit callout

content/concepts/preferences.mdx
```diff
@@ -5,33 +5,66 @@ tags: ["recipients", "conditions", "prefs"]
...
 
-In the next section, we'll walk through this evaluation model in more detail and outline how to set preferences via the Knock preferences API.
+- [At the environment level](#environment-level-preference-defaults), as the default for all recipients in a given environment.
+- [At the tenant level](#per-tenant-preference-defaults), as the default for all recipients when a `tenant` is applied to a workflow trigger.
+- [At the level of an individual recipient](#preference-sets).
```
JEverhart383, line 53:
```suggestion
- [At the level recipient level](#preference-sets), which can be a [User](concepts/users) or an [Object](concepts/objects).
```

cellomatt: For this one, I left off the explicit user/object reference because we have it both at the top of the page and a callout when you click through to the linked section. Happy to add it here if you think it's better to be more explicit, I just didn't want to get pedantic about it

content/concepts/preferences.mdx
```diff
@@ -5,33 +5,66 @@ tags: ["recipients", "conditions", "prefs"]
...
+  <Step title="You trigger the workflow to send notifications">
+    When the event that you want to send notifications about occurs, you'll
+    [trigger](/send-notifications/triggering-workflows) your workflow.
+  </Step>
+  <Step title="Notifications are sent">
+    The workflow runs for every `recipient` in the triggering API call, sending
+    them a notification on each channel step that you've configured in the
+    workflow.
+  </Step>
+</Steps>
```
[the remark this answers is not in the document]
cellomatt: Yeah, I don't disagree! If other folks are ok with chopping this out, I think we can save a lot of space

content/concepts/preferences.mdx
JEverhart383: I think I was just after something that worked with the parallel structure here in the 'At the *** level`, so we can remove the extra details

cellomatt: Got it, makes sense! ty

cellomatt:
> I'll come back through with some additional suggestions, but just wanted to offer my 2 cents on your open question. I wonder if we could lift the code samples from '/managing-recipients/setting-preferences' to '/concepts/preferences' to better support some of those examples and collapse those pages. Then I think the 'client integration' JS code from '/managing-recipients/setting-preferences' would actually make more sense to live in '/in-app-ui/javascript/preferences'

Maybe another possible approach is turning the existing Preferences concept page here into a very brief summary of what Prefs are, then breaking all of these concepts up into multiple pages under a "Preferences" section within **Managing Recipients** (where Setting Preferences currently is)?


image

content/concepts/preferences.mdx
rachael-t: I'm onboard and agree with this 👍

content/concepts/preferences.mdx
```diff
@@ -60,17 +95,15 @@ Here's a JSON example of a given recipient's workflow preferences.
...
-
-To set a category for a given workflow, go to that workflow's page in the dashboard, click the "..." menu, and select "Manage workflow". From there you'll be able to add categories. **Note:** workflow categories are **case sensitive**.
+There are cases when you'll want to enable your recipients to set a preference for a category of workflows; for example, all of your workflows related to `direct` notifications. In these cases, rather than having to surface a preference control for each of those notification types, you can use `categories` to set the preference for multiple workflows at once.
 
-A workflow can belong to multiple categories. In the instance that it does, and conflicting preferences are set across those categories, the workflow's preference will default to false.
+To set a `category` for a given workflow, go to that workflow's page in the dashboard, click the "..." menu, and select "Manage workflow". From there you'll be able to add categories. **Note: workflow categories are case sensitive**.
```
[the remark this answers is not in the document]
JEverhart383: Yeah, my 2 cents here is that we should move the language describing 'how to create workflow categories' to this page and just link to that heading from here: https://docs.knock.app/concepts/workflows#workflow-categories

There's no how to language on the actual workflow page unless I'm missing something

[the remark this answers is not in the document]
cellomatt: I'm all for saving space, I think we could do this with a "learn how to set a workflow category here" link

content/concepts/preferences.mdx
```diff
@@ -188,19 +234,19 @@ If your product just gives users a single set of notification preferences to con
   }
 />
 
-In some cases you may want to set an additional preference set for a given user that pertains to a specific tenant within your product. (A well-known example of this: in Slack, your notification preferences are set _per Slack workspace_, not as a global preference set that applies across all of your Slack workspaces.)
+In some cases you may want to set an additional preference set for a given user that pertains to a specific `tenant` within your product. (A well-known example of this: in Slack, your notification preferences are set _per Slack workspace_, not as global preferences that apply across all of your Slack workspaces.)
 
-To create a preference set scoped to a particular [Tenant](/send-and-manage-data/tenants), you'll set the `id` of the preference set equal to `tenant.id` when you create the preference set for your user.
+To create a preference set scoped to a particular [Tenant](/send-and-manage-data/tenants), you'll set the `id` of the recipient's preference set equal to the `tenant.id`.
```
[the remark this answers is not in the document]
cellomatt: lol fair!

rachael-t:
> > I'll come back through with some additional suggestions, but just wanted to offer my 2 cents on your open question. I wonder if we could lift the code samples from '/managing-recipients/setting-preferences' to '/concepts/preferences' to better support some of those examples and collapse those pages. Then I think the 'client integration' JS code from '/managing-recipients/setting-preferences' would actually make more sense to live in '/in-app-ui/javascript/preferences'
> 
> Maybe another possible approach is turning the existing Preferences concept page here into a very brief summary of what Prefs are, then breaking all of these concepts up into multiple pages under a "Preferences" section within **Managing Recipients** (where Setting preferences currently is)?
> 
> image

I don't have strong feelings either way, but I think we could play around with exploring what doing this would look like on a different PR to see if splitting them out like that feels better or if we think it chunks it apart too much

JEverhart383 (approved):
All solid improvements IMO

rachael-t (approved):
This looks great! Let's ship these updates! 🚀

PR #480, opened 2024-05-01
fix: prevent `Endpoints` from overflowing container

cellomatt:
### Description

This PR fixes a UI bug in the `Endpoints` component which was allowing endpoints that exceeded a certain character length to overflow their container. The update adds a horizontal scroll bar for endpoints that are too long (mimicking the code block components) and a bit of margin on the endpoints themselves so the scrollable section has equal padding on either side.

See https://docs-git-mk-endpoints-component-knocklabs.vercel.app/reference#channel-data for an example

### Screenshots

image

image

image

### DIFF
--- a/components/Endpoints.tsx
+++ b/components/Endpoints.tsx
@@ -8,7 +8,9 @@ const Endpoints = ({ children, title = "Endpoints" }) => (
         {title}
       </span>
     </div>
-    <div className="m-0 py-2 px-4 dark:bg-gray-900 overflow-x-auto">{children}</div>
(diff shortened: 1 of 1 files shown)


cellomatt:
Hmmm, I just found a separate issue related to how we use the individual `Endpoint` component 🤔 This exists on the `main` branch and isn't affected by this PR, but maybe we should do some additional thinking about best fix here 
image

rachael-t (approved):
Nice catch!

I think we can merge this change and then think about how we want to address the endpoint/code block overlap in another PR

PR #470, opened 2024-04-18
chore: add accordion groups to all FAQ sections

cellomatt:
### Description

This PR places all of our "Frequently asked questions" into accordion groups. It also makes fixes for typos and outdated information ("trigger conditions," anyone?) that I found along the way.

Apologies for the massive PR, it got bigger than I expected and I was already too deep 😅 

### Screenshots

**Note:** For FAQs that had only a single question, I opted to default the accordion to "open".

image

image

PR #483, opened 2024-05-02
feat: add one signal push integration docs

connorlindsey:
### Description
Adds documentation for OneSignal push integration

### Tasks
[KNO-5890](https://linear.app/knock/issue/KNO-5890/%5Bonesignal%5D-docs)

### DIFF
--- a/content/integrations/push/one-signal.mdx
+++ b/content/integrations/push/one-signal.mdx
@@ -1,0 +1,72 @@
+---
+title: How to send push notifications using OneSignal
+description: How to send mobile push notifications with OneSignal and Knock.
+tags: ["react native", "ios", "android", "push"]
+section: Integrations
+layout: integrations
+---
+
+In this guide we'll walk through how to configure a OneSignal provider in Knock to send push notifications. This guide assumes that you've already created a OneSignal channel in the Knock dashboard.
+
+Your OneSignal channel expects that you are using Knock to author Push notification templates. Those templates are then passed to OneSignal as content.
+
+## How to configure OneSignal Push with Knock
+
+**Note**: this guide assumes that you have already setup OneSignal with push certificates and everything needed in order to start sending push notifications. You should also have already implemented the OneSignal mobile SDK within your application.
+
+To setup OneSignal Push with Knock you will need:
+
+- Your OneSignal API Key for sending notifications
+- To select a mode of operation between using `player_ids` (deprecated) or `external_ids` (recommended)
+
+## Using OneSignal Push with Knock
+
+### Using `player_id`
+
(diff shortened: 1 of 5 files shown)


content/integrations/push/one-signal.mdx
```diff
@@ -0,0 +1,73 @@
...
+You can retrieve the player_id from OneSignal once a user has been registered, either by using the OneSignal mobile SDK, or by using a Webhook from OneSignal.
+
+<MultiLangCodeBlock
+  snippet="users.setChannelData-one-signal"
+  title="Setting a player_id as channel data for a user"
+/>
+
+### Using `external_id` (recommended)
+
+If you’re using an external id to reference your OneSignal users within Knock, you do not need to supply any additional data to initiate a request. We will automatically pass your Knock user id to OneSignal as the `external_id`.
```
rachael-t, line 38: Since using `external_id` is recommended, I think we should list that first.

content/integrations/push/one-signal.mdx
```diff
@@ -0,0 +1,73 @@
...
+
+## How to configure OneSignal Push with Knock
+
+**Note**: this guide assumes that you have already setup OneSignal with push certificates and everything needed in order to start sending push notifications. You should also have already implemented the OneSignal mobile SDK within your application.
+
+To setup OneSignal Push with Knock you will need:
+
+- Your OneSignal App ID
+- Your OneSignal API Key for sending notifications
+- To select a mode of operation between using `player_ids` (deprecated) or `external_ids` (recommended)
```
rachael-t, line 21: I'm wondering why we mention that `player_ids` is deprecated but then also provide the steps below on how to use that? I guess the (deprecated) is throwing me off a little

rachael-t (approved):
Looks great! I added comments about some confusion I had about `player_ids` and I also have a few small nits (but they aren't blocking to getting these docs out).

1. I think changing the note on line 15 to a callout component would help draw attention
2. Matt recently moved all FAQs to use the accordion components (PR #470 ) - I'd suggest doing so here just for consistency
3. We should also add OneSignal to the overview page (`/integrations/push/overview#supported-providers`)

Again, I don't think those are blocking, so feel free to merge in and I can go ahead and get those 3 changes up separately.

content/integrations/push/one-signal.mdx
connorlindsey: It's deprecated by OneSignal and will be replaced by other methods (external_id being the one we're going to support). I'll clarify that.

rachael-t: ah gotcha! thank you!

PR #485, opened 2024-05-04
fix: update example-app filename to plural form 

rachael-t:
### Description

Updates example-app filename + page slug to plural form since it now lists out multiple example apps that we have available. I noticed this discrepancy when I clicked [on the link under "What is Knock?"](https://github.com/knocklabs/docs/blob/main/content/getting-started/what-is-knock.mdx?plain=1#L219) and received a 404 since that referenced "example-apps" and decided that we should update the filename and page slug for accuracy.


cellomatt (approved):
🥳

PR #487, opened 2024-05-06
chore: add example table definition

rachael-t:
### Description

Adds an example of what the table definitions from Prequel are for customers to be able to review prior to completing the setup via the magic link. It also fixes a breadcrumb issue on the `/data-sync` page.

https://docs-git-rt-add-schema-definitions-knocklabs.vercel.app/integrations/extensions/data-sync

Let me know if there's any information I should exclude from the example (i.e., varying lengths)

### DIFF
--- a/content/integrations/extensions/data-sync.mdx
+++ b/content/integrations/extensions/data-sync.mdx
@@ -24,9 +24,7 @@ Once the information is filled out, Prequel will provide you with a table defini
     Please note that the table name will vary based on the name of the schema provided when filling out the form.
     ```markdown
         CREATE TABLE <name of schema>.messages (
-          message_id character varying(65535) NOT NULL ENCODE raw
-          distkey
-        ,
+          message_id character varying(65535) NOT NULL ENCODE raw distkey,
             account_id character varying(65535) ENCODE lzo,
             environment_id character varying(65535) ENCODE lzo,
             environment_name character varying(65535) ENCODE lzo,
@@ -57,9 +55,7 @@ Once the information is filled out, Prequel will provide you with a table defini
             has_been_interacted bigint ENCODE az64,
             has_been_archived bigint ENCODE az64,
             PRIMARY KEY (message_id)
-        ) DISTSTYLE KEY
-        SORTKEY
-          (message_id, updated_at);
+        ) DISTSTYLE KEY SORTKEY (message_id, updated_at);
     ```
   </Accordion>
 </AccordionGroup>
--- a/content/integrations/extensions/data-sync.mdx
+++ b/content/integrations/extensions/data-sync.mdx
@@ -17,10 +17,16 @@ First, you'll have to give Knock your database type and host destination address
 
 Next, Knock generates a magic link you'll fill out. See what information you'll need for your particular database [here](https://docs.prequel.co/reference/post_import-magic-links).
 
-Once the information is filled out, Prequel will provide you with a table definition (example provided below). Additionally, you can run a test connection to make sure it's wired up. If it's successful, you can save it and it will proceed with the backfill transfer.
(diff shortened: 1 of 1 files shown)


content/integrations/extensions/data-sync.mdx
```diff
@@ -1,22 +1,64 @@
...
 ## Setup
 
 First, you'll have to give Knock your database type and host destination address for your data warehouse. The databases supported as destinations for Knock are [those listed on Prequel](https://docs.prequel.co/docs/destinations-overview).
 
 Next, Knock generates a magic link you'll fill out. See what information you'll need for your particular database [here](https://docs.prequel.co/reference/post_import-magic-links).
 
-Once the information is filled out, you can run a test connection to make sure it's wired up. If it's successful, you can save it and it will proceeed with the backfill transfer.
+Once the information is filled out, Prequel will provide you with a table definition (example provided below). Additionally, you can run a test connection to make sure it's wired up. If it's successful, you can save it and it will proceed with the backfill transfer.
+
+<AccordionGroup>
```
cellomatt, line 22: I feel like this might fit better under the "Available data" section? This is essentially the same info, just formatted as a SQL query for creating the table, right? So maybe we fit both this table definition and the **Message table structure** within one `AccordionGroup`?

rachael-t: Yup it sure is and that makes more sense! Just pushed up updated changes - let me know your thoughts!

content/integrations/extensions/data-sync.mdx
```diff
@@ -1,184 +1,226 @@
...
+        CREATE TABLE <name of schema>.messages (
+          message_id character varying(65535) NOT NULL ENCODE raw distkey,
+            account_id character varying(65535) ENCODE lzo,
+            environment_id character varying(65535) ENCODE lzo,
+            environment_name character varying(65535) ENCODE lzo,
+            environment_slug character varying(65535) ENCODE lzo,
+            channel_id character varying(65535) ENCODE lzo,
+            channel_name character varying(65535) ENCODE lzo,
+            channel_key character varying(65535) ENCODE lzo,
+            channel_type character varying(65535) ENCODE lzo,
```
rachael-t, line 41: I have no idea why starting with `channel_type` the number of tabs/spacing changes. I've checked, and I can't find a lingers empty space. They should all have the same number of tabs 🤔 
Screenshot 2024-05-06 at 1 29 46 PM

content/integrations/extensions/data-sync.mdx
rachael-t: Hmm I will say moving the "Message table structure" into an accordion removes the anchor tag (realizing this as I've sent https://docs.knock.app/integrations/extensions/data-sync#messages-table-structure to customers as they are looking to set up the data warehouse connector). Maybe this is worth diving into outside of this PR but when we move things into accordions, we also lose the anchor tags and being able to directly link customers to the information they need.

cellomatt: Yeah that's a good point. It also is a bit tough to know which one to click on just from looking at the accordion titles. I think it's probably fine as-is now, but if you wanted to experiment with one more change maybe you could give each accordion a `description` (like the demo one at docs.knock.app/playground) and move some of the language from the paragraph above into there 🤷 anyway, not point in belaboring it and I think it's fine as-is

cellomatt (approved):

content/integrations/extensions/data-sync.mdx
rachael-t: haha that's totally it! how annoying! 😅

rachael-t: This is resolved. Specifying `markdown` (or leaving it blank, which I assume defaults to markdown) for syntax highlighting creates this behavior—updated to `sql`.

PR #488, opened 2024-05-07
chore: homepage cards

cellomatt:
### Description

Reformat docs homepage to use cards instead of listing a full page index.

As part of this update, I also made some tiny modifications to the `Card` component (which we may or may not like):
- Slight font-size increase to the `title` on each card (this will impact `SdkCard` instances as well)
- moved padding for the card onto the inner `<a>` tag so that the entire card is clickable

### Questions

Do we like this approach? Is it ok that mobile users will lose the ability to navigate to specific pages?

### DIFF
--- a/components/Card.tsx
+++ b/components/Card.tsx
@@ -11,7 +11,7 @@ type Props = {
 
 const CardGroup = ({ children, cols = 2 }) => (
   <div
-    className={clsx(`grid gap-2 auto-rows-[minmax(140px,auto)]`, {
+    className={clsx(`grid gap-2`, {
       "grid-cols-2": cols == 2,
       "grid-cols-3": cols == 3,
     })}
--- a/components/Card.tsx
+++ b/components/Card.tsx
@@ -11,7 +11,7 @@ type Props = {
 
 const CardGroup = ({ children, cols = 2 }) => (
   <div
-    className={clsx(`grid gap-2`, {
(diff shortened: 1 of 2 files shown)


JEverhart383:
Here's a screenshot of our mobile/desktop traffic across the docs site, and mobile traffic to that page is still about 25% of total traffic, so I think you're asking a good question about removing some navigation affordance for those users.
Screenshot 2024-05-07 at 9 23 40 AM

The comment I came here to leave is that I'd love to see this page, and the cards in particular, become like a hot spot for the most likely tasks based on analytics data, rather than just enumerate what's in the sidebar. But, now looking at the mobile data, that idea is at odds with what I outlined above.

cellomatt:
> The comment I came here to leave is that I'd love to see this page, and the cards in particular, become like a hot spot for the most likely tasks based on analytics data, rather than just enumerate what's in the sidebar. But, now looking at the mobile data, that idea is at odds with what I outlined above.

Theoretically, if we move the sidebar into a hamburger for smaller screen widths we would take care of that navigation issue which would free us up to be more flexible with the rest of the page. I love the idea of featuring common recipes, etc and not just the main topics

cellomatt:
Closing this since it was mainly conceptual and we've taken care of the short-term things mentioned above (mobile menu and fixing clickable area on the SDK card)!

PR #494, opened 2024-05-09
chore: update dashboard navigation instructions (part 2)

cellomatt:
### Description

This is Part 2 of a broader update begun in #492. Because I opened this as a branch that builds off that PR, it probably makes sense to review/merge them in order (there are "only" 22 new files updated here)!

The main thrust of this PR is to update mentions of navigating in the dashboard to be consistent in their styling: 
- a **Page** in the dashboard is bolded, and a series of page navigations will look like **Page One** > **Page Two**
- a "Button" label (or other clickable entity) will have quotes 

Along the way, I'm also fixing information which is outdated or no longer fully accurate.


samseely (approved):
💯

PR #495, opened 2024-05-10
chore: use import w/ node exs

JEverhart383:
### Description

Did a find/replace on all the of the import/require statements for the Node.js multilang code examples


cellomatt (approved):
:icon:

PR #496, opened 2024-05-10
chore: update dashboard navigation instructions (part 3)

cellomatt:
### Description

Part 3 (final).

The main thrust of this PR is to update mentions of navigating in the dashboard to be consistent in their styling: 
- a **Page** in the dashboard is bolded, and a series of page navigations will look like **Page One** > **Page Two**
- a "Button" label (or other clickable entity) will have quotes 

I also fixed as many instances as I could find where we improperly put punctuation outside of quotation marks, i.e. `",` => `,"` and `".` => `."`


content/mapi.mdx
```diff
@@ -1317,7 +1317,7 @@ Note: This immediately enables or disables a workflow in a given environment wit
   <Attribute
     name="environment"
     type="string"
-    description='The environment slug. (Defaults to "development".)'
+    description='The environment slug. (Defaults to "development.")'
```
cjbell, line 1320: All of these read so weirdly to me, in British English we do the previous convention which I guess is why it's rampant in the docs 🙃

cjbell (approved):

content/mapi.mdx
cellomatt: Interesting!!

PR #507, opened 2024-05-16
Fix bugs in Go docs snippets

vlymar:
### Description
Brings the docs improvements from the go sdk readme into our docs site (https://github.com/knocklabs/knock-go/pull/33)

Also fixes some other compilation errors in our go docs snippets. Most of my focus in this PR was on snippets related to workflow triggering, we should also audit the remaining go snippets.

### DIFF
--- a/data/code/workflows/trigger-with-branding-tenant.ts
+++ b/data/code/workflows/trigger-with-branding-tenant.ts
@@ -79,14 +79,19 @@ $client->workflows()->trigger('reservation-reminder-email', [
 ctx := context.Background()
 knockClient, _ := knock.NewClient(knock.WithAccessToken("sk_12345"))
 
-result, _ := knockClient.Workflows.Trigger(ctx, &knock.TriggerWorkflowRequest{
+request := &knock.TriggerWorkflowRequest{
   Workflow:   "reservation-reminder-email",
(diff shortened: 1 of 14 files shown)


cjbell (approved):

data/code/workflows/trigger-with-branding-tenant.ts
```diff
@@ -79,14 +79,19 @@ $client->workflows()->trigger('reservation-reminder-email', [
...
-  Recipients: reservationGuestIds,
   Tenant: "black-lodge"
-  })
+}
+
+for _, r := range reservationGuestIds {
+  request.AddRecipientByID(r)
+}
+
+result, _ := knockClient.Workflows.Trigger(ctx, req, nil)
```
cellomatt, line 94:
```suggestion
result, _ := knockClient.Workflows.Trigger(ctx, request, nil)
```
I think either this or we should refactor `request` above to match other examples? Also realizing I'm late to the party but still noting here so we don't forget

vlymar: good 👁️, I'll open a followup PR in which I'll update all exampls to use `request` rather than `req` to fix this and be consistent everywhere else

PR #509, opened 2024-05-16
chore: add usenotifstore docs and remove zustand usage

cjbell:
### Description

* Adds new docs for new `useNotificationStore` hook
* Removes all `zustand` usage

### DIFF
--- a/content/sdks/react/reference.mdx
+++ b/content/sdks/react/reference.mdx
@@ -558,9 +558,36 @@ Renders a notification bell icon, with a badge showing the number of unseen item
 
 Creates an authenticated Knock client.
 
-**Returns**: `Knock` instance, authenticated against the user
+#### Params
 
-**Example**:
+<Attributes>
+  <Attribute
+    name="apiKey*"
+    type="string"
+    description="The public API key for the Knock environment"
+  />
+  <Attribute
+    name="userId"
+    type="string"
+    description="The userId of the user to authenticate with Knock."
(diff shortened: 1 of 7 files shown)


content/sdks/react/reference.mdx
```diff
@@ -635,6 +683,48 @@ const MyComponent = () => {
 };
 ```
 
+### `useNotificationStore`
+
+Given a `Feed` will return reactive access to the `FeedStateStore`. Can optionally accept a `selector` to retrieve a fine-grained slice of state from the store.
```
connorlindsey, line 688: Would it be helpful to include a note that this is using Zustand under the hood? I don't think it's necessary but it could provide more detail on writing selectors

content/sdks/react/reference.mdx
```diff
@@ -601,16 +628,38 @@ Exposed under `KnockI18nProvider` child components.
...
+
+#### Example
 
 ```js
 import {
   useAuthenticatedKnockClient,
   useNotifications,
+  useNotificationStore,
 } from "@knocklabs/react";
-import create from "zustand";
```
connorlindsey, line 613: Do we want to include an migration guide?

cjbell: One neat thing is that it will continue to work for current customers, but yes could absolutely include in the repo.

connorlindsey (approved):

PR #516, opened 2024-05-22
chore: clarify current lack of support for Object feeds

cellomatt:
### Description

This PR updates the Objects concept page to make it clearer that support for an Object recipient of an in-app feed is not currently available but on the roadmap. Current state seems to indicate that it's possible in the list, and only mentions that it isn't yet available in the roadmap callout below.


samseely (approved):
Thanks for this one, Matt!

PR #526, opened 2024-05-28
Sts translations refresh

samseely:
A refresh on translations docs as we add support for t tags. 

@connorlindsey A couple questions: 
* Do we need to call out PO files in here in some way? 
* On code example under "Commit your workflow", it looks like we're using autogenerated id for translation key but I believe this should be the translation string itself?
image

### DIFF
--- a/content/concepts/translations.mdx
+++ b/content/concepts/translations.mdx
@@ -34,9 +34,9 @@ section: Concepts
 
 To get started, enable translations for your account. Go to “Settings” under your account name in the left sidebar and click “Enable translations”.
 
-Next you'll need to set a default locale. Knock uses the default locale when it can't find a translation for a given recipient’s locale.
+Next you'll need to set a default `locale`. Knock uses the default `locale` when it can't find a translation for a given recipient’s `locale`.
 
-Once you've set your default locale, you should see a new “Translations” page under “Developers” in the sidebar. This is where you’ll be working with your translations.
+Once you've set your default `locale`, you should see a new “Translations” page under “Developers” in the sidebar. This is where you’ll be working with your translations.
 
 ## Basic usage
 
@@ -64,7 +64,7 @@ Once you have those translations created for the `en` and `fr` locales, you can 
 <p> {{ "OrderReady" | t }} </p>
 ```
 
-Your users must have a `locale` property set for the helper to find translations in their locale. You can set a user's `locale` with the [identify endpoint](/reference#identify-user).
+Your users must have a `locale` property set for the helper to find translations in their locale, otherwise Knock will use the default locale. You can set a user's `locale` with the [identify endpoint](/reference#identify-user).
 
 ## Translation methods: filter vs. tag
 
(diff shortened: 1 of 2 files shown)


connorlindsey:
@samseely Good question on PO files. I think if anything, we can add a small note saying we support pushing/pulling PO files using the CLI/mAPI but since the rest of our system (including the UI) uses our object format, I'm hesitant to call it out too much since we don't really have full PO support. 

And yeah, we aren't auto-generating keys anymore so that should be the content (e.g. `Have your received your order?`)

samseely:
> @samseely Good question on PO files. I think if anything, we can add a small note saying we support pushing/pulling PO files using the CLI/mAPI but since the rest of our system (including the UI) uses our object format, I'm hesitant to call it out too much since we don't really have full PO support.

@connorlindsey On the PO stuff, can you take first swing at drafting language for that? Think you'll be best person to draft since you're closest to it. Feel free to just push straight up to this PR or add as a suggestion on review. I'm good to just include as a section under CLI H2.

content/concepts/translations.mdx
```diff
@@ -114,19 +110,15 @@ You can and probably will often use variables inside of your translations. Let
...
 ```
 
-Where “actor” and “likedPhoto” are variables that get passed into your data payload and have a name and title key, respectively.
-
 ### Pluralization
 
-Translations sometimes need specialized pluralization rules. When you pass the `count` variable to a translation, it will know to look for a special set of keys within your translation to match it to. Those keys are `zero`, `one`, and `other`. You don’t need to reference these in the template, if you pass the `count` variable it will evaluate it and choose one for you.
-
-For example:
+Translations support pluralization rules. When you pass the `count` variable to a translation, it looks for pluralization keys in your translation. Those keys are `zero`, `one`, and `other`. You don’t need to reference these in the template, if you pass the `count` variable it will evaluate it and choose one for you.
```
JEverhart383, line 121:
```suggestion
Translations support pluralization rules. When you pass the `count` variable to a translation, it looks for pluralization keys in your translation. Those keys are `zero`, `one`, and `other`. You don’t need to reference these in the template. If you pass the `count` variable, it will evaluate it and choose one for you.
```

content/concepts/translations.mdx
```diff
@@ -151,17 +143,19 @@ And you would just need to pass your `count` variable for it to choose one of th
 
 ### Other filters in combination
 
-You can still use other filters in combination with `t` but you’ll use them **\***after**\*** you use the `t` filter. For example, to titlecase a translation:
+You can still use other filters in combination with `t` but you’ll use them **\***after**\*** you use the `t` filter.
```
JEverhart383, line 146:
```suggestion
You can still use other filters in combination with `t` but you’ll use them **after** you use the `t` filter.
```
assuming bold emphasis here

content/concepts/translations.mdx
```diff
@@ -151,17 +143,19 @@ And you would just need to pass your `count` variable for it to choose one of th
...
+For example, to titlecase a translation:
 
 ```json title="Message template editor"
 <p> {{ "congratulationsMessage" | t | titlecase }} </p>
 ```
 
 ### Namespaced translations
 
-You’ll use namespaced translations just as you would a translation without a namespace, but you’ll use a special way to access it in the message template. Make sure you have the same namespace across your locales if you use them so that the message template will be able to access all of the content in them. Here’s an example of using a namespaced translation in message template.
+When you create a translation, you can supply an optional “namespace”. The namespace helps organize translations of the same locale so you can keep similar concepts together. Below you'll see examples of how to reference namespaced translations from your message templates.
```
JEverhart383, line 156:
```suggestion
When you create a translation, you can supply an optional “namespace.” The namespace helps organize translations of the same locale so you can keep similar concepts together. Below you'll see examples of how to reference namespaced translations from your message templates.
```

content/mapi.mdx
```diff
@@ -1706,7 +1706,7 @@ An email layout
 
 [Translations](/send-and-manage-data/translations) support localization in Knock. They hold the translated content for a given locale, which you can reference in your message templates with the `t` Liquid function filter.
 
-You can retrieve, update, and create translations as well as listing all in a given [environment](/send-and-manage-data/environments). Translations are identified by their locale code + an optional namespace.
+You can retrieve, update, and create translations as well as list all translations in a given [environment](/send-and-manage-data/environments). Translations are identified by their locale code + an optional namespace.
```
JEverhart383, line 1709:
```suggestion
You can retrieve, update, and create translations as well as list all translations in a given [environment](/concepts/environments). Translations are identified by their locale code + an optional namespace.
```

Update old URL

PR #528, opened 2024-05-31
feature: add video and card

JEverhart383:
### Description
Screenshot 2024-05-31 at 9 20 30 AM

This adds a new icon option to the SDK card and uses it to call out videos where we have them. I think embedding most videos it too disruptive, but in some cases where the concept is complex a link might not be suggestive enough, so this is a nice-looking middle ground. 

Open to feedback here though cause it's still somewhat in your face

### DIFF
--- a/components/SdkCard.tsx
+++ b/components/SdkCard.tsx
@@ -9,6 +9,7 @@ import {
   FaSwift,
   FaAngular,
   FaVuejs,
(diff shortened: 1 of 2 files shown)


samseely (approved):
Nice, I like the card treatment on this one. Not too intrusive but gives them a nice path out to video.

PR #529, opened 2024-05-31
fix: hot reloading and hydration errors when editing docs locally

connorlindsey:
### Description
- Hot reloads docs when editing content in the `/content` folder
- Fixes a hydration error on the translations page caused by nested anchor tags. I clicked through every page and didn't see any other hydration errors pop up. 
- Fixes an issue with the support overlay where the sidebar was rendering on top of the overlay where it should be behind it. 

### Tasks
[KNO-6060](https://linear.app/knock/issue/KNO-6060/%5Bdocs%5D-fix-hydration-errors-when-editing-docs-locally)

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -1,3 +1,10 @@
+const path = require("path");
+
+// Refresh the site when content changes
+const withRemoteRefresh = require("next-remote-refresh")({
+  paths: [path.resolve(__dirname, "content")],
+});
+
 /** @type {import('next').NextConfig} */
 const nextConfig = {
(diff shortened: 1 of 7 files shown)


next.config.js
```diff
@@ -1,3 +1,10 @@
+const path = require("path");
+
+// Refresh the site when content changes
+const withRemoteRefresh = require("next-remote-refresh")({
+  paths: [path.resolve(__dirname, "content")],
+});
```
cjbell, line 6: YES YES YES

cjbell (approved):
❤️

PR #536, opened 2024-06-06
chore: add docs for new actionable in-app notifs

cjbell:
### Description

- Adds reference for types for new block types
- Removes inline actions callout
- Adds reference for new `onNotificationButtonClick` events

### Tasks

KNO-6136

### DIFF
--- a/content/in-app-ui/api-overview.mdx
+++ b/content/in-app-ui/api-overview.mdx
@@ -281,6 +281,6 @@ In-app notification messages (`FeedItem`) are "actionable" and can be one of thr
 - **Single-action**: indicates that the notification cell has a single action button that can be clicked.
 - **Multi-action**: indicates that the notification cell has both a primary and secondary set of actions that can be clicked.
 
-Working with actionable notifications is straightforward. If you're using out-of-the-box Knock UI components, we'll handle the rendering of actionable notifications for you. 
+Working with actionable notifications is straightforward. If you're using out-of-the-box Knock UI components, we'll handle the rendering of actionable notifications for you.
 
 If you're building a custom feed implementation, you can check the `blocks` attribute for the presence of a `ButtonSetContentBlock` and render the `buttons` contained within the block accordingly.
--- a/content/in-app-ui/api-overview.mdx
+++ b/content/in-app-ui/api-overview.mdx
@@ -281,6 +281,6 @@ In-app notification messages (`FeedItem`) are "actionable" and can be one of thr
 - **Single-action**: indicates that the notification cell has a single action button that can be clicked.
 - **Multi-action**: indicates that the notification cell has both a primary and secondary set of actions that can be clicked.
 
-Working with actionable notifications is straightforward. When rendering the in-app message, you can check the `blocks` attribute for the presence of a `ButtonSetContentBlock` and render the `buttons` contained within the block accordingly.
+Working with actionable notifications is straightforward. If you're using out-of-the-box Knock UI components, we'll handle the rendering of actionable notifications for you. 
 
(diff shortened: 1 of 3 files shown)


content/in-app-ui/api-overview.mdx
```diff
@@ -198,3 +272,15 @@ It's also possible to override the response of the in-app API to hide sensitive
...
+
+In-app notification messages (`FeedItem`) are "actionable" and can be one of three types:
+
+- **Standard**: indicates that the entire cell is potentially actionable (e.g., clicking anywhere on the cell will trigger an action).
+- **Single-action**: indicates that the notification cell has a single action button that can be clicked.
+- **Multi-action**: indicates that the notification cell has both a primary and secondary set of actions that can be clicked.
+
+Working with actionable notifications is straightforward. When rendering the in-app message, you can check the `blocks` attribute for the presence of a `ButtonSetContentBlock` and render the `buttons` contained within the block accordingly.
+
+If you're using out-of-the-box Knock UI components, we'll handle the rendering of actionable notifications for you.
```
JEverhart383, line 286:
```suggestion
```

content/in-app-ui/api-overview.mdx
```diff
@@ -198,3 +272,15 @@ It's also possible to override the response of the in-app API to hide sensitive
...
+
+## Working with actionable notifications
+
+In-app notification messages (`FeedItem`) are "actionable" and can be one of three types:
+
+- **Standard**: indicates that the entire cell is potentially actionable (e.g., clicking anywhere on the cell will trigger an action).
+- **Single-action**: indicates that the notification cell has a single action button that can be clicked.
+- **Multi-action**: indicates that the notification cell has both a primary and secondary set of actions that can be clicked.
+
+Working with actionable notifications is straightforward. When rendering the in-app message, you can check the `blocks` attribute for the presence of a `ButtonSetContentBlock` and render the `buttons` contained within the block accordingly.
```
JEverhart383, line 284:
```suggestion
Working with actionable notifications is straightforward. If you're using out-of-the-box Knock UI components, we'll handle the rendering of actionable notifications for you. 

If you're building a custom feed implementation, you can check the `blocks` attribute for the presence of a `ButtonSetContentBlock` and render the `buttons` contained within the block accordingly.
```
Maybe lead with base case of we handle for you and call out the custom alternative, in which case they'd already need to be handling the other content block

content/sdks/react/reference.mdx
```diff
@@ -400,6 +400,11 @@ A provider to inject translations into components.
     type="function"
     description="A custom function to be invoked when a notification cell is clicked."
   />
+  <Attribute
```
JEverhart383, line 403: Likely too late, but I feel like `onActionButtonClick` might have been a better name - I think we're likely to see some confusion on when to use this vs. `onButtonClick`. With this release is the other method still relevant?

JEverhart383 (approved):
LGTM

content/sdks/react/reference.mdx
cjbell: I actually disagree - this prop is on the `NotificationFeed` and the `onButtonClick` is on the `NotificationCell`. To keep things consistent I chose `onNotificationButtonClick` to be in line with `onNotificationClick` which is meant only for a whole cell click.

JEverhart383: Ahh, fair. I think I misunderstood the purpose of one vs. the other. Retracted. If I understand these examples correctly, `onButtonClick` and `onNotificationButtonClick` are essentially the same thing, just specified at different levels of the component tree. And since many people may not actually customize a `NotificationCell` they just use the top level prop
Screenshot 2024-06-06 at 4 08 00 PM

PR #543, opened 2024-06-12
chore: try to highlight web push

JEverhart383:
### Description
Try to do a better job of calling out web push support here for FCM. After looking at the existing docs, the steps are almost the same, so just tried a few small tweaks to make it more apparent that you could do web push here and added Algolia tags to make it show up in search


rachael-t (approved):
I think this looks great! Thanks for actioning this 👍

PR #548, opened 2024-06-21
chore: add new message api filtering docs

cjbell:
### Description

Adds new message API filtering methods

### DIFF
--- a/content/reference.mdx
+++ b/content/reference.mdx
@@ -874,12 +874,22 @@ Retrieve a paginated list of messages for a user. Will return most recent messag
   <Attribute
     name="status"
     type="string[] (optional)"
-    description="One or more of `queued`, `sent`, `delivered`, `delivery_attempted`, `undelivered`, `bounced`, `not_sent`. Limits results to messages with the given delivery status(es)"
+    description="One or more of `queued`, `sent`, `delivered`, `delivery_attempted`, `undelivered`, `bounced`, `not_sent`. Limits results to messages with the given delivery status(es)."
+  />
+  <Attribute
+    name="engagement_status"
+    type="string[] (optional)"
+    description="One or more of `read`, `unread`, `seen`, `unseen`. Limits results to messages with the given engagement status(es)."
   />
   <Attribute
     name="channel_id"
     type="string (optional)"
     description="Limits the results to only items that belong to the channel"
+  />
+  <Attribute
+    name="messge_ids[]"
+    type="string[] (optional)"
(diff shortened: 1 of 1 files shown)


content/reference.mdx
```diff
@@ -874,13 +874,23 @@ Retrieve a paginated list of messages for a user. Will return most recent messag
...
   />
   <Attribute
     name="channel_id"
     type="string (optional)"
     description="Limits the results to only items that belong to the channel"
   />
+  <Attribute
+    name="messge_ids[]"
+    type="string[] (optional)"
+    description="Limits the results to only the message ids given (max 50). Note: when using this option, all other filters will be"
```
JEverhart383, line 892: Looks like this is missing a part of the sentence, 'all other filters will be...'

content/reference.mdx
```diff
@@ -3913,13 +3973,23 @@ Returns a paginated list of messages for an object. Will return newest messages
...
```
JEverhart383, line 3991: same as above

content/reference.mdx
```diff
@@ -2816,11 +2846,21 @@ Returns a paginated list of messages.
...
+  />
   <Attribute
     name="channel_id"
     type="string (optional)"
     description="Limits the results to only items that belong to the channel"
   />
+  <Attribute
+    name="messge_ids[]"
+    type="string[] (optional)"
+    description="Limits the results to only the message ids given (max 50). Note: when using this option, all other filters will be"
```
JEverhart383, line 2862: same as above

JEverhart383 (approved):
I think there's just one missing piece of a description, otherwise LGTM

content/reference.mdx
```diff
@@ -889,7 +889,7 @@ Retrieve a paginated list of messages for a user. Will return most recent messag
   <Attribute
     name="messge_ids[]"
     type="string[] (optional)"
-    description="Limits the results to only the message ids given (max 50). Note: when using this option, all other filters will be"
+    description="Limits the results to only the message ids given (max 50). Note: when using this option, messages returned will also match any other filters applied."
```
JEverhart383, line 892:
```suggestion
    description="Limits the results to only the message ids given (max 50). Note: when using this option, the results will be subject to the any other filters applied."
```
Hmmm...the `also` and `match` here make this feel additive when it will really only ever be fewer results b/c of other filters. Suggested one option above, but here's another: `when using this option, any other filters on the query will still be applied and further limit the result set.`

rachael-t: Good call out! I personally think the suggested change helps to clarify this.

rachael-t (approved):

PR #561, opened 2024-07-09
chore: add courier migration guide

cellomatt:
### Description

This PR introduces a new Guide on migrating from Courier to Knock. It includes the relevant navigation update, and also updates the `section` on the "Moving to production" guide that was previously moved in the nav but not updated.

In addition to review of written content, it would be helpful to get feedback (if any) on any potential visual improvements as well.

https://docs-yg1ymevhd-knocklabs.vercel.app/guides/migrate-from-courier

### DIFF
--- a/content/guides/migrate-from-courier.mdx
+++ b/content/guides/migrate-from-courier.mdx
@@ -90,7 +90,7 @@ Knock offers APIs and developer tools that make a migration smooth and efficient
   emoji="🚸"
   text={
     <>
-      <strong>Inline recipient identification:</strong> While the following
+      <strong>Inline recipient identification.</strong> While the following
       steps outline a suggested order for migrating individual resource types
       into Knock based on your existing Courier integration, it’s helpful to
       note that Knock also supports{" "}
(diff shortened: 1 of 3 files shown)


rachael-t (approved):
This looks great! It's thorough, but I don't feel like it's _too_ in-depth. While there is a lot of content, it works well for me visually. I especially like that migration steps are added to the page nav. 

Kudos to you, Matt, for a great job on this! 🎉

content/guides/migrate-from-courier.mdx
```diff
@@ -0,0 +1,244 @@
...
+            <Endpoint method="GET" path="https://api.courier.com/notifications" />
+          </a>
+          <a href="https://docs.knock.app/mapi#workflows-update" target="_blank" style={{textDecoration: 'none'}}>
+            <Endpoint method="PUT" path="https://control.knock.app/v1/workflows/:workflow_key" />
+          </a>
+      </Accordion>
+    </AccordionGroup>
+
+  </Step>
+  <Step title="Import Translations files">
```
samseely, line 140: Nit: on all these step titles let's use sentence case. we're doing it on some items "Migrate users" but not others "Add Tenants". Good to be consistent. 

I also wondered whether we should use sentence case on H2s under mapping section of this doc, but given those are first-order concepts, I think makes sense to capitalize them.

content/guides/migrate-from-courier.mdx
```diff
@@ -0,0 +1,244 @@
...
+
+- A [Management API](/mapi) that allows you to work programmatically with the resources that you can also create directly in your Knock dashboard (like Workflows and their associated message templates, email [Layouts](/integrations/email/layouts), and Translations).
+- A command line interface ([Knock CLI](/developer-tools/knock-cli)) that wraps the Management API, allowing you to work with your dashboard resources from the command line.
+- [Bulk endpoints](/reference#bulk-endpoints) that allow you to upsert large amounts of data in a single API request (more on specific endpoints below).
+
+<Callout
+  emoji="🚸"
+  text={
+    <>
+      <strong>Inline recipient identification:</strong> While the following
```
samseely, line 93:
```suggestion
      <strong>Inline recipient identification.</strong> While the following
```
Nit: on the callouts we usually end the bold part with a period instead of colon when it's introducing a new concept (exception is for small things like "Note"). Would be good to add to docs style guide!

samseely (approved):
Fantastic job on this @cellomatt! Couple small nits but otherwise g2g!

content/guides/migrate-from-courier.mdx
cellomatt: @samseely yeah, I capitalized "Translations" here to refer to the first-order concept. It's tough to know which to use where, since all of these concepts could also be used as just their english words 😂 if this feels weirdly inconsistent though, I can update

cellomatt: Maybe just "Import translation files"

PR #562, opened 2024-07-12
fix: engagement_statuses typo for messages endpoint

rachael-t:
### Description

Updates `engagement_status` query params for the messages endpoints to `engagement_statuses`


content/reference.mdx
```diff
@@ -877,7 +877,7 @@ Retrieve a paginated list of messages for a user. Will return most recent messag
     description="One or more of `queued`, `sent`, `delivered`, `delivery_attempted`, `undelivered`, `bounced`, `not_sent`. Limits results to messages with the given delivery status(es)."
   />
   <Attribute
-    name="engagement_status"
+    name="engagement_statuses"
```
cjbell, line 880: This was actually correct as documented and designed to match the status filter

rachael-t: Ah yeah, I see that now. Sorry about that. I had tested this, and it wasn't working, but I now see the error I was making.

PR #566, opened 2024-07-16
chore: add info about actor being excluded from recipients

cellomatt:
### Description

This PR updates the list of effects of including an `actor` on a trigger to include the fact that an `actor` is never notified as a `recipient`. Right now, this information is rather buried in the Subscriptions FAQ.

https://docs-a86l84q42-knocklabs.vercel.app/send-notifications/triggering-workflows#attributing-the-action-to-a-user-or-object


rachael-t (approved):
🙌

PR #570, opened 2024-07-18
fix: correct EventData attributes for a `message.link_clicked` event

cellomatt:
### Description

We currently list the extra `EventData` param that is included with a `message.link_clicked` event incorrectly. It should be a `url`, not `link_clicked`.

https://docs-fmj18zbrl-knocklabs.vercel.app/developer-tools/outbound-webhooks/event-types#messagelink_clicked

I checked this in the codebase and also confirmed with a webhook test:
```
{
.... 
"event_data": {
    "__typename": "EventData",
    "url": "https://www.google.com"
  },
  "type": "message.link_clicked"
}
```


JEverhart383 (approved):
nice catch

PR #572, opened 2024-07-19
chore: remove complex js -> node replacement logic and update code examples

cellomatt:
### Description

Because the `MultiLangCodeBlock` block now references the code examples to populate its switcher, we don't need to have the complex logic to make sure syntax highlighting and the language stored in localstorage reference `javascript` examples for the Node SDK; we already have both `javascript` (for Vanilla JS) and `node` as registered languages using `javascript` highlighting, so it's much less confusing to just reference one or the other in the code examples depending on whether it's for the Node SDK or vanilla JS.

This PR removes all of the complicated logic around the above, and updates all of the code examples that reference the Node SDK to use `node` instead of `javascript`.

- The playground page has an updated code block with both `node` and `javascript` to show the difference: https://docs-kcs6p0uhr-knocklabs.vercel.app/playground
- The API reference should appear unchanged: https://docs-kcs6p0uhr-knocklabs.vercel.app/reference

### DIFF
--- a/components/MultiLangCodeBlock.tsx
+++ b/components/MultiLangCodeBlock.tsx
@@ -122,6 +122,7 @@ const DEFAULT_ORDER: SupportedLanguage[] = [
   "curl",
   "shell",
   "node",
+  "javascript",
   "jsweb",
   "ruby",
   "python",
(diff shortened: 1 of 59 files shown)


connorlindsey (approved):
Looks good! Thanks for updating every example

PR #574, opened 2024-07-22
chore: add cURL examples to API reference

cellomatt:
### Description
This PR adds cURL examples to our API reference example code documentation. It also updates some of the existing code examples from our SDKs to make them more consistent (i.e. keep variable names the same for the same use case across languages).

https://docs-dt6mzgbkm-knocklabs.vercel.app/reference

### Testing
To ensure accuracy, I've tested each of these cURL examples by using the "copy" button on the code example in the reference, pasting the copied code into my terminal, then manually updating the API key and any other values (like `channel_id` or `message_id`) that need to be changed in order to get a valid response. These are all working as expected.

### To-do
- To keep this already-large PR as brief as possible, I held off on updating some of the code examples that are used in other areas of the documentation (i.e. preferences, inline identification, and channel data examples). I'll create a follow-up PR to add cURL to those examples.
- There are some endpoints in the reference that currently feature only the request body (no code examples). In some cases, this is because we don't have corresponding SDK examples. I think it might be nice to feature these as cURL requests rather than just a JSON request body, as I think it helps to be a bit more explicit; however, we may want to hold off on those until we also have generated SDKs, as the work might be faster at that point. I did add one of these ([List objects](https://docs-dt6mzgbkm-knocklabs.vercel.app/reference#list-objects)).


samseely:
This is a pretty epic update, Matt! Great work. I'm approving but probably one that Jeff or Chris should take a closer look at before we merge.

samseely (approved):

PR #581, opened 2024-07-31
chore(kno-6592): add docs on log truncation

meryldakin:
### Description
Tells users when and why we truncate logs. I didn't add specific numbers on what size we truncate at, but we can if we think that would be helpful. 

### Tasks
[kno-6592](https://linear.app/knock/issue/KNO-6592/[docs]-docs-on-log-truncation)


connorlindsey (approved):
No strong opinion on if we want to include the exact length where we truncate logs, maybe worth including if it's helpful for debugging anything, but probably fine to leave it out.

PR #586, opened 2024-08-06
chore: update notes on webhook page to use the callout component

rachael-t:
### Description

Updates the **Notes:** on the webhook overview page to use the `<Callout>` component instead to draw the user's attention better.  

https://docs-git-rt-clean-up-webhook-notes-knocklabs.vercel.app/integrations/webhook/overview


cellomatt (approved):
🥳

PR #591, opened 2024-08-13
feat(pla-330): expand docs on data retention policies + enforcement

bceskavich:
### Description

- Adds a new page covering data retention policies
- Adds links back to the data retention page from relevant guides
- Adds a section on retention enforcement in the V1 API reference + info to each endpoint affected by retention rules

### Tasks

[PLA-330: Write docs on message retention enforcement](https://linear.app/knock/issue/PLA-330/write-docs-on-message-retention-enforcement)

### DIFF
--- a/content/manage-your-account/data-retention.mdx
+++ b/content/manage-your-account/data-retention.mdx
@@ -1,0 +1,55 @@
+---
+title: Data retention
+description: How Knock enforces data retention policies on your account.
+tags: ["data", "retention"]
+section: Manage your account
+---
+
+Knock applies a retention policy to some of the data within your account. Once data exceeds its retention period, it will no longer be accessible in the Knock Dashboard, the Management API, or the V1 API. Knock will eventually prune data that has aged out of its retention period.
+
+## Data retention policies
+
+<Callout
+  emoji="⚠️"
+  text={
+    <>
+      <span className="font-bold">
+        Custom retention policies are not currently supported.
+      </span>{" "}
+      If custom retention windows are critical to your usage of Knock, please
+      use the "Help" dropdown at the top of this page to contact support.
+    </>
+  }
+/>
+
+Some Knock data is subject to a retention policy based on your account plan. These policies are:
+
+- **Enterprise plans**: 90 days
+- **All other plans**: 30 days
+
+Knock applies this plan-based retention policy to the following data:
+
+- [Audit logs](/manage-your-account/audit-logs)
+- [Message log data](/concepts/messages)
+- [Outbound webhook delivery logs](/developer-tools/outbound-webhooks/overview#reading-webhook-delivery-logs)
+- [Source event and action logs](/integrations/sources/overview#logs)
+- [Workflow run logs](/send-notifications/debugging-workflows)
+
+Additionally, Knock applies a universal, 30-day retention policy to the following data:
+
+- [API logs](/developer-tools/api-logs)
+
+## Message log data retention details
+
+Message log data subject to your account's retention policy includes both the message log itself and associated metadata, including:
+
+- Message contents
+- Message events
+- Delivery logs
+- Activities
+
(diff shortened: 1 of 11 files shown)


content/manage-your-account/data-retention.mdx
```diff
@@ -0,0 +1,55 @@
...
+  text={
+    <>
+      <span className="font-bold">
+        Custom retention policies are not currently supported.
+      </span>{" "}
+      If custom retention windows are critical to your usage of Knock, please
+      use the "Help" dropdown at the top of this page to contact support.
+    </>
+  }
+/>
```
bceskavich, line 23: For context: @cjbell and I discussed this as something we could add later on, but at the moment it's not prioritized.

content/manage-your-account/data-retention.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+- Message contents
+- Message events
+- Delivery logs
+- Activities
+
+However, this does not mean that the associated notification has been completely removed from Knock's systems. Your recipients can still interact with these older notifications, and you can still take some specific actions on them.
+
+1. **In-app feed notifications are available indefinitely.** Even once your message log data has expired, you can still fetch the associated `FeedItem` data from the [Show feed endpoint](/reference#get-feed) to deliver in-app notifications to your recipients.
+2. **Message status updates are allowed indefinitely.** You can still call the [Update message status](/reference#update-message-status), [Remove message status](/reference#undo-message-status), and [Batch change message status](/reference#batch-update-message-status) endpoints to update the status of an older message, either directly or via a Knock SDK. These status updates will still generate new message events and trigger your [outbound webhooks](/developer-tools/outbound-webhooks/overview), even though you will not be able to view the data in the Knock Dashboard any longer.
```
bceskavich, line 54: Some further context: when updating the status of an old message, we'll return a `PrunedMessage` entity in the response, which contain only:

- The original message ID
- The channel ID
- The recipient info
- Statuses

I don't explicitly note that here, nor do I mention it in the V1 API reference, since we don't currently mention the response to the update status calls at all. I'm wondering if we should add that in?

brentjanderson: Customers will find out here or when they try to implement it that the response is different. I think it's worth calling out in the docs. For me, getting an API response that is not the same as what the docs describes is not a good sort of "surprise". Even a footnote is enough to get my attention that "This will be different" so I can account for it accordingly.

content/manage-your-account/data-retention.mdx
```diff
@@ -0,0 +1,55 @@
...
+Message log data subject to your account's retention policy includes both the message log itself and associated metadata, including:
+
+- Message contents
+- Message events
+- Delivery logs
+- Activities
+
+However, this does not mean that the associated notification has been completely removed from Knock's systems. Your recipients can still interact with these older notifications, and you can still take some specific actions on them.
+
+1. **In-app feed notifications are available indefinitely.** Even once your message log data has expired, you can still fetch the associated `FeedItem` data from the [Show feed endpoint](/reference#get-feed) to deliver in-app notifications to your recipients.
```
brentjanderson, line 53: > Indefinitely

That's a very long time. I'm of the opinion that _all_ data needs a lifecycle that includes when it will be fully retired from the system.

> [!WARNING]
> The following is scope creep.

I'm wondering if we need additional options for customers to control retention on messages themselves? Each plan tier/custom enterprise agreement can specify what the upper bound can be for message retention, and we enforce the TTL accordingly. Once it hits that limit, we drop the message entirely.

Why is that not possible? And for us to support an indefinite message use case, customers need to pay us indefinitely for the retention of that message?

I get that's probably beyond the scope of this PR & current line of work, but I think we need to at least know how we can get there and plan on making that part of the product at some point.

Otherwise, indefinite retention is a long time, and we will have to have this conversation at some point in the future.

PR #598, opened 2024-08-16
Update documentation link

matonny:
### Description
Hi there, I believe the documentation on overriding Expo payloads through the template editor might be linking to wrong expo docs. Instead of the Push API docs it links to client side types and so there's a bit of a mismatch in terms of fields that can be specified :))


cellomatt:
Thanks @matonny!! Including this in #599 🙏

PR #605, opened 2024-08-22
feat: add docs for workflow trigger data validation

cjbell:
### Description

Adds docs for workflow trigger validation

### Tasks

KNO-6818

### DIFF
--- a/content/send-notifications/triggering-workflows.mdx
+++ b/content/send-notifications/triggering-workflows.mdx
@@ -235,7 +235,7 @@ If you specify a workflow trigger frequency of "Once per recipient", then you ca
   emoji="🚧"
   text={
     <>
-      This is a pre-release feature. Please reach out to support@knock.app if
+      This is a pre-release feature. Please reach out to [support@knock.app](mailto:support@knock.app) if
       you'd like to try it out.
     </>
   }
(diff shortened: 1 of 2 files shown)


content/send-notifications/triggering-workflows.mdx
```diff
@@ -228,3 +228,35 @@ In some situations, you may want to limit the number of times that a recipient c
...
 
 If you specify a workflow trigger frequency of "Once per recipient", then you can also optionally choose to include the tenant in the frequency control. When set, your workflow will only trigger once per-recipient, per-tenant.
+
+## Validating workflow trigger data
+
+<Callout
+  emoji="🚧"
+  text={
+    <>
+      This is a pre-release feature. Please reach out to support@knock.app if
```
rachael-t, line 238:
```suggestion
      This is a pre-release feature. Please reach out to [support@knock.app](mailto:support@knock.app) if
```

rachael-t (approved):
Small suggestion, but not blocking 😄

content/send-notifications/triggering-workflows.mdx
```diff
@@ -228,3 +228,35 @@ In some situations, you may want to limit the number of times that a recipient c
...
+      you'd like to try it out.
+    </>
+  }
+/>
+
+Given how critical the data is that's passed to your workflow via the trigger API call, you may wish to validate the trigger data sent to Knock before a workflow is run. This is where the trigger data validation feature comes in to play.
+
+Trigger data validation lets you specify a [JSON schema](https://json-schema.org/) that describes the shape of the data that's passed to your workflow. If the data doesn't match the schema, then the trigger endpoint will return a `422` error with details of where the data failed validation.
```
JEverhart383, line 246: Just a few thoughts here:

1. If we wanted a deeper link into the JSON schema site, this page gets to an example pretty quickly: https://json-schema.org/learn/getting-started-step-by-step
2. Is it worth mentioning we don't require `$schema` or `$id` keywords or annotations?
3. Maybe also worth calling out that the expected `type` will always be `object` at the top level?
This is one of those things I've only had limited exposure to, so just Qs I had reading through that getting started link, but I think most people will be able to run with that example. Excited to get this out there, very cool feature 🙌

JEverhart383 (approved):
🔥 just some thoughts, none blocking

PR #607, opened 2024-08-27
fix: correct broken links and add links to endpoints in reference "overview" sections

cellomatt:
### Description

Our `Endpoint` component has a `withLink` prop that allows a given listed endpoint to link to the place in the reference where it is documented.

We apply `withLink` inconsistently right now; it's only used in a few sections on the reference, but it's actually very convenient if you're looking at the "Overview" of any of the sidebar sections.

This PR:
- adds `withLink` where it was missing
- corrects some links that were broken because the `name` on the endpoint didn't match the anchor tag for the relevant documentation
- updates the way that we reference the path parameter for loading an in-app feed (we use `:feed_id` in most places but also referenced `:in_app_feed_channel_id` in a few spots).
- removes the internal-use `GET /users/:user_id/feeds/:feed_id/settings` endpoint that was listed but not documented

https://docs-gpn7f3jqi-knocklabs.vercel.app/reference

### TODO

As I was going through this, I found 2 endpoints that we list in the Overview sections but do not actually document below. I didn't add `withLink` to them because there's nowhere to anchor them to, but I'll add them in a followup PR:

- `GET /users/:user_id/preferences` (list all user preferences)
- `GET /objects/:collection/:id/preferences` (list all object preferences)

### Screenshots

image

### DIFF
--- a/content/reference.mdx
+++ b/content/reference.mdx
@@ -488,11 +488,13 @@ A Workflow orchestrates the delivery of messages to your end users. When you con
     name="trigger-workflow"
     method="POST"
     path="/workflows/:key/trigger"
+    withLink
   />
   <Endpoint
     name="cancel-workflow"
     method="POST"
     path="/workflows/:key/cancel"
+    withLink
   />
 </Endpoints>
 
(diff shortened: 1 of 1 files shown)


samseely (approved):
🔥🔥🔥 Great docs improvement PR, Matt. Thanks for taking initiative to level this up and finding some missing API refs along the way

PR #615, opened 2024-09-09
feat: introduce `data` in template variables

cjbell:
### Description

- Move language to use `data.` to reference variables
- Fix template vars table

### Tasks

KNO-6914


cellomatt (approved):
lgtm!

PR #623, opened 2024-09-23
chore: fix mapi formatting

cellomatt:
### Description

This PR does a few things to improve formatting of the mAPI reference:

1. Use the `<Section>` component where we had `<div>` tags to improve spacing and ensure consistency with the content columns
2. Add the various sub-definitions of Workflows into their own section in the sidebar menu to make them navigable and easier to find
3. Link all of the Knock `Types` in various `<Attribute>` components to their definitions

https://docs-a6x3z4evv-knocklabs.vercel.app/mapi

### Question

I noticed that the list of pages under "Workflow definitions" doesn't open and highlight the "active" page as expected, which I think is because the component really intends for these to be separate pages under a shared `/path`. I'm looking into that, but if it's something quick that someone else is already aware of re: the way that the component functions, let me know 😃 I personally think it's an acceptable compromise for organization as it is.

### Screenshots
Before:
image

After: 
image

### DIFF
--- a/content/mapi.mdx
+++ b/content/mapi.mdx
@@ -124,10 +124,8 @@ You can retrieve, update, or create a workflow as well as list all workflows in 
 </ExampleColumn>
 </Section>
 
-<div className="flex flex-col md:flex-row w-full py-16" id="workflows-object">
+<Section title="Workflow definition" slug="workflows-object">
 <ContentColumn>
-
-## Workflow definition
 
 ### Attributes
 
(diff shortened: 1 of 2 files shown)


rachael-t (approved):
Thank you so much for doing this 🙌 

I wonder if we should rename the sub-definitions (i.e., drop "definition")? It feels a little repetitive and wordy on the side nav, but I also see the value within the page itself.

samseely (approved):
Great improvements, Matt! 💯 Let's ship it

PR #627, opened 2024-09-25
chore: update mapi reference

cellomatt:
### Description

This PR makes the following updates to the MAPI reference:

- Add attributes on the `Workflow` definition for `sha`, `trigger_frequency`, and `trigger_data_json_schema`
- Add a `conditions` object to the example workflow object
- Update some non-typed `object` references to their defined `Types`
- Add `typeSlug` links with additional context to many attributes throughout the reference, in order to link back to their definitions

https://docs-9pc4eysuz-knocklabs.vercel.app/mapi

### For consideration

I used the `typeSlug` to link a few things (like the `conditions` and `trigger_data_json_schema`) back to their definitions, even though they aren't "typed" per se; they're just an `object`. I think this is an acceptable use for it since it takes you to the place where you can learn more about the expected format of that `object`, but just wanted to flag that this is an atypical use for that in case there's a better way to approach.

### DIFF
--- a/content/mapi.mdx
+++ b/content/mapi.mdx
@@ -222,9 +222,9 @@ You can retrieve, update, or create a workflow as well as list all workflows in 
   "conditions": {
     "all": [
       {
-        "argument": "admin",
+        "variable": "recipient.role",
         "operator": "equal_to",
-        "variable": "recipient.role"
+        "argument": "admin"
       }
     ]
   },
--- a/content/mapi.mdx
+++ b/content/mapi.mdx
@@ -198,7 +198,7 @@ You can retrieve, update, or create a workflow as well as list all workflows in 
     name="trigger_frequency"
     type="string"
     typeSlug="/send-notifications/triggering-workflows#controlling-workflow-trigger-frequency"
-    description="The frequency at which the workflow should be triggered. One of: ”once_per_recipient”, ”every_trigger”. Defaults to ”every_trigger”."
+    description="The frequency at which the workflow should be triggered. One of: ”once_per_recipient”, ”once_per_recipient_per_tenant”, ”every_trigger”. Defaults to ”every_trigger”."
   />
   <Attribute
     name="updated_at"
--- a/content/mapi.mdx
+++ b/content/mapi.mdx
@@ -141,6 +141,12 @@ You can retrieve, update, or create a workflow as well as list all workflows in 
     description="A list of categories that the workflow belongs to."
   />
   <Attribute
+    name="conditions"
+    type="object (optional)"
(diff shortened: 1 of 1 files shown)


rachael-t (approved):
Nice updates! I think the `typeSlug` usage is fine, and I actually find it helpful.

JEverhart383:
Agreed, `typeSlug` link is kind of nice, everything else LGTM

content/mapi.mdx
```diff
@@ -165,6 +171,11 @@ You can retrieve, update, or create a workflow as well as list all workflows in
     type="string"
     description="A name for the workflow. Must be at maximum 255 characters in length."
   />
+  <Attribute
+    name="sha"
+    type="string (read-only)"
+    description="The SHA hash of the workflow's most-recent commit."
```
thomaswhyyou, line 177:
```suggestion
    description="The SHA hash of the workflow data."
```
It does not necessarily have to be a committed change/data.

content/mapi.mdx
```diff
@@ -178,9 +189,16 @@ You can retrieve, update, or create a workflow as well as list all workflows in
...
+    name="trigger_data_json_schema"
+    type="object (optional)"
+    typeSlug="/developer-tools/validating-trigger-data#example-schema"
+    description="A JSON schema for the expected structure of the workflow trigger's data payload. Used to validate trigger requests."
+  />
+  <Attribute
+    name="trigger_frequency"
+    type="string"
+    typeSlug="/send-notifications/triggering-workflows#controlling-workflow-trigger-frequency"
+    description="The frequency at which the workflow should be triggered. One of: ”once_per_recipient”, ”every_trigger”. Defaults to ”every_trigger”."
```
thomaswhyyou, line 201: There is a third option `once_per_recipient_per_tenant` I believe - but should confirm with @natperpepaj @meryldakin.

thomaswhyyou (approved):
A couple small things, but lgtm! Also like the type slug links, thank you Matt!!

content/mapi.mdx
cellomatt: Thanks! Is it more accurate to say the workflow version, then?

content/mapi.mdx
natperpepaj: yea that's correct. once_per_recipient_per_tenant means that the workflow will trigger once for every unique user <> tenant pairing

cellomatt: Ok got it, ty! Will update before merging

natperpepaj: ty!

PR #629, opened 2024-09-28
chore: add channel overrides to mAPI

cellomatt:
### Description

This PR adds a `ChannelSettings` type to the mAPI reference, so that it can be referenced in the `channel_overrides` attribute on a `ChannelStep`. It also links the attribute for push channel JSON overrides (which are at the template level) back to the place where its documented.

I also snuck in a mention of Partials at the top of the reference where we list the resources you can manage with mAPI.

**Note:** While both Push and Webhook channel types have `ChannelSettings` attributes in our product, they aren't currently relevant to `channel_overrides` and so I opted not to document them here. If we add endpoints to allow channel integration management via API at some point in the future, we'll want to add these other attributes so that we can reference the `ChannelSettings` type in those endpoints.

https://docs-git-mk-mapi-part-two-knocklabs.vercel.app/mapi#channelsettings-definition

### DIFF
--- a/content/mapi.mdx
+++ b/content/mapi.mdx
@@ -683,7 +683,7 @@ Contains all settings that are applied to a channel integration. Used to overrid
   />
 </Attributes>
 
-{/* 
+{/\*
 Not relevant to channel_overrides. Should be added if we add endpoints to configure channel integrations.
 
 #### Push channel settings attributes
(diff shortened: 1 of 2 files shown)


thomaswhyyou (approved):
Very nice, great stuff @cellomatt! 👏👏

PR #632, opened 2024-10-03
feat(pla-443): add docs on trigger data filtering + warehouse export

bceskavich:
### Description

- Adds a top section to the reference on how Knock manages trigger data for filtering
- Updates endpoints that accept `trigger_data` filters to point to this reference
- Adds note on trigger data exported to warehouse connections

### Tasks

[PLA-443: Write docs on `searchable_trigger_data` filter usage + warehouse export](https://linear.app/knock/issue/PLA-443/write-docs-on-searchable-trigger-data-filter-usage-warehouse-export)

### DIFF
--- a/components/Attributes.tsx
+++ b/components/Attributes.tsx
@@ -4,10 +4,21 @@ const Attributes = ({ children }) => (
   </div>
 );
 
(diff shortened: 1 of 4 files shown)


rachael-t (approved):
Thank you for these updates! 🙌

PR #635, opened 2024-10-08
update reference guide for liquid

JEverhart383:
### Description
Adds a reference item for `date` filter having `timezone` option and adds rec. to `timezone` filter when using with `date`
Screenshot 2024-10-08 at 10 08 06 AM


content/designing-workflows/template-editor/reference-liquid-helpers.mdx
```diff
@@ -21,19 +21,20 @@ The Knock template editor uses Liquid syntax for control flow and variable decla
...
+| ------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | --------------------------------------------------------------------------------- |
+| `timezone`         | Takes an ISO 8601 timestamp and returns it in the [IANA tz database timezone](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) provided. You can use the built-in `timestamp` [template variable](/designing-workflows/template-editor/variables) to reference the current datetime. When formatting with the `date` filter, we recommend using the `timezone` option on the `date` filter instead if displaying the timezone abbreviation with `%Z`. | <code>{"{{ timestamp \| timezone: \"America/New_York\"}"}}</code>                 |
```
natperpepaj, line 26: non-blocking suggestion: some people might not be reaching for the %Z format on their datetime so this might just be added confusion but fine with whatever you think is best

```suggestion
| `timezone`         | Takes an ISO 8601 timestamp and returns it in the [IANA tz database timezone](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) provided. You can use the built-in `timestamp` [template variable](/designing-workflows/template-editor/variables) to reference the current datetime. When formatting with the `date` filter, we recommend using the `timezone` option on the `date` filter instead. | <code>{"{{ timestamp \| timezone: \"America/New_York\"}"}}</code>                 |
```

natperpepaj (approved):

PR #638, opened 2024-10-08
chore: auth token requirement clarification

rachael-t:
### Description

We received feedback that the current wording of "optionally an auth token for production environments" made it seem unclear when it is needed. This update clarifies that we stronglu recommend using it for production environments, and if enhanced security mode is enabled, it will be required. 

Also updates a link that was broken.

https://docs-git-rt-update-rn-sdk-token-auth-knocklabs.vercel.app/sdks/react-native/quick-start#configuration


content/sdks/react-native/quick-start.mdx
```diff
@@ -21,7 +21,23 @@ To get started, you will need the following:
...
 1. A public API key (found in the Knock dashboard)
-2. A user ID, and optionally an auth token for production environments
+2. A user ID and an auth token
+
+   <Callout
+     emoji="💡"
+     text={
+       <>
+         Auth tokens are strongly recommended for production environments and
+         required when enhanced security mode is enabled. For more information,
```
cellomatt, line 31:
```suggestion
         are required when enhanced security mode is enabled. For more information,
```

cellomatt (approved):
nit on wording but this is great!

PR #641, opened 2024-10-10
feat(kno-7125): add relative delay documentation

natperpepaj:
### Description
Add relative delay step type documentation for feature [Computed Delay Windows](https://www.notion.so/knockapp/Computed-delay-windows-fac767ee652048c4addc94207a3bc68a)

### Tasks
[KNO-7125](https://linear.app/knock/issue/KNO-7125/[docs]-add-docs-for-relative-delay)

### Screenshots

TODO

### DIFF
--- a/content/designing-workflows/delay-function.mdx
+++ b/content/designing-workflows/delay-function.mdx
@@ -121,7 +121,7 @@ You can set a relative delay to compute a delay time from a set interval and a d
 
 When configuring a relative delay, you'll specify:
 
-- The set interval (e.g., 1 day, 2 hours, 30 minutes)
+- The delay interval (e.g., 1 day, 2 hours, 30 minutes)
 - Whether the delay should occur before or after the dynamic timestamp
 - The `key` for the dynamic timestamp (which can come from your trigger data, recipient data, or other sources)
 
--- a/content/designing-workflows/delay-function.mdx
+++ b/content/designing-workflows/delay-function.mdx
@@ -5,7 +5,7 @@ tags: ["steps", "delays", "wait", "functions"]
 section: Designing workflows
 ---
 
-A delay function does just what it sounds like: it delays the execution of the workflow for some amount of time, then proceeds to the next step. There are three types of delays we support in Knock today: "wait for fixed interval", "wait for a dynamic duration", and  "wait until a relative timestamp."
+A delay function does just what it sounds like: it delays the execution of the workflow for some amount of time, then proceeds to the next step. There are three types of delays we support in Knock today: "wait for fixed interval", "wait for a dynamic period", and  "wait until a relative timestamp."
 
 ## Wait for a fixed interval
 
--- a/content/designing-workflows/delay-function.mdx
+++ b/content/designing-workflows/delay-function.mdx
@@ -129,7 +129,7 @@ As in the dynamic duration section above, the key specified must be available to
 
 Relative interval delay functions are helpful for various scenarios, including:
 
-- Appointment reminders: Send a notification 1 day before an appointment time
+- Appointment reminders: send a notification one day before an appointment time
 - Follow-up reminders: Send a follow-up message 2 hours after an event
 
 ## Using workflow cancellation with delays
--- a/content/designing-workflows/delay-function.mdx
+++ b/content/designing-workflows/delay-function.mdx
@@ -121,7 +121,7 @@ You can set a relative delay to compute a delay time from a set interval and a d
 
 When configuring a relative delay, you'll specify:
 - The set interval (e.g., 1 day, 2 hours, 30 minutes)
-- Whether the delay should occur before or after the dynamic timestamp
+- Whether the delay should occur before or after the dynamic duration
 - The `key` for the dynamic duration (which can come from your trigger data, recipient data, or other sources)
 
 As in the dynamic duration section above, the key specified must be available to be resolved. If the key is missing or resolves to an invalid value, a corresponding error will be logged on the workflow run, and the delay will be **skipped**.
--- a/content/designing-workflows/delay-function.mdx
+++ b/content/designing-workflows/delay-function.mdx
@@ -123,7 +123,7 @@ When configuring a relative delay, you'll specify:
 
 - The set interval (e.g., 1 day, 2 hours, 30 minutes)
 - Whether the delay should occur before or after the dynamic timestamp
-- The `key` for the dynamic duration (which can come from your trigger data, recipient data, or other sources)
+- The `key` for the dynamic timestamp (which can come from your trigger data, recipient data, or other sources)
 
 As in the dynamic duration section above, the key specified must be available to be resolved. If the key is missing or resolves to an invalid value, a corresponding error will be logged on the workflow run, and the delay will be **skipped**.
 
--- a/content/designing-workflows/delay-function.mdx
+++ b/content/designing-workflows/delay-function.mdx
@@ -115,6 +115,21 @@ When the key specified is missing or resolves to an invalid value, a correspondi
   </Accordion>
 </AccordionGroup>
 
+## Wait for a relative duration
+
+You can set a relative delay to compute a delay time from a set interval and a dynamic duration. This means you can set a delay that occurs a set amount of time before or after a provided dynamic timestamp. This allows the creation of flexible and context-aware delay periods in your workflows.
+
+When configuring a relative delay, you'll specify:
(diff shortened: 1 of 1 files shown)


content/designing-workflows/delay-function.mdx
```diff
@@ -115,6 +115,23 @@ When the key specified is missing or resolves to an invalid value, a correspondi
...
 </AccordionGroup>
 
+## Wait for a relative duration
+
+You can set a relative delay to compute a delay time from a set interval and a dynamic duration. This means you can set a delay that occurs a set amount of time before or after a provided dynamic timestamp. This allows the creation of flexible and context-aware delay periods in your workflows.
+
+When configuring a relative delay, you'll specify:
+
+- The set interval (e.g., 1 day, 2 hours, 30 minutes)
+- Whether the delay should occur before or after the dynamic duration
```
cellomatt, line 125:
```suggestion
- Whether the delay should occur before or after the dynamic timestamp
```
I wonder if this would be clearer? I know we use "dynamic duration" in the section above, but that's when the delay itself is dynamic based on a dynamically-provided explicit timestamp, where as now we're saying the duration is the interval that is calculated based on that explicit timestamp. I think it's confusing to say both "interval" and "duration" but refer to two separate things

content/designing-workflows/delay-function.mdx
```diff
@@ -115,6 +115,23 @@ When the key specified is missing or resolves to an invalid value, a correspondi
...
+- The `key` for the dynamic duration (which can come from your trigger data, recipient data, or other sources)
```
cellomatt, line 126:
```suggestion
- The `key` for the dynamic timestamp (which can come from your trigger data, recipient data, or other sources)
```
Same question as above. Or maybe "dynamic duration's timestamp" 🤔

cellomatt (approved):
looks good with a couple questions/suggestions. Maybe we need to rethink the way we use the term "delay duration" altogether in light of this feature?

content/designing-workflows/delay-function.mdx
natperpepaj: i think you're right was back and forth with this since we call the above step type a dynamic duration, however the more obvious use case for this feature is to use a timestamp

natperpepaj: going to keep this comment section unresolved while we get feedback

content/designing-workflows/delay-function.mdx
```diff
@@ -115,6 +115,23 @@ When the key specified is missing or resolves to an invalid value, a correspondi
   </Accordion>
 </AccordionGroup>
 
+## Wait for a relative duration
+
+You can set a relative delay to compute a delay time from a set interval and a dynamic duration. This means you can set a delay that occurs a set amount of time before or after a provided dynamic timestamp. This allows the creation of flexible and context-aware delay periods in your workflows.
```
cellomatt, line 120:
```suggestion
You can set a relative delay to compute a delay time from a set interval and a dynamic timestamp. This means you can set a delay that occurs a set amount of time before or after a provided dynamic timestamp. This allows the creation of flexible and context-aware delay periods in your workflows.
```
Just realized we use it here too, but then later in the sentence say "timestamp." My personal preference is to just use "timestamp" consistently

content/designing-workflows/delay-function.mdx
```diff
@@ -115,6 +115,23 @@ When the key specified is missing or resolves to an invalid value, a correspondi
   </Accordion>
 </AccordionGroup>
 
+## Wait for a relative duration
```
samseely, line 118:
```suggestion
## Wait until a relative timestamp
```

My thinking here is language maps more closely to UI and to what this functionally doing. "Relative duration" kind of confuses me (could say it is same as "dynamic duration"?) but "wait until a relative timestamp" feels more specific to what this is doing.

content/designing-workflows/delay-function.mdx
samseely, line 120:
```suggestion
You can use our relative delay to wait some time before or after a timestamp that you provide in your workflow payload. Relative delays are great for use cases such as reminders, where you want to wait until a day before a due date to remind your users.
```

I'm finding current language pretty hard to parse. Let's aim to simplify and get this to a tangible use case as quickly as possible.

content/designing-workflows/delay-function.mdx
```diff
@@ -115,6 +115,23 @@ When the key specified is missing or resolves to an invalid value, a correspondi
...
+
+You can set a relative delay to compute a delay time from a set interval and a dynamic duration. This means you can set a delay that occurs a set amount of time before or after a provided dynamic timestamp. This allows the creation of flexible and context-aware delay periods in your workflows.
+
+When configuring a relative delay, you'll specify:
+
+- The set interval (e.g., 1 day, 2 hours, 30 minutes)
```
samseely, line 124:
```suggestion
- The delay interval (e.g., 1 day, 2 hours, 30 minutes)
```
Just to make consistent with bullet that comes after it

content/designing-workflows/delay-function.mdx
```diff
@@ -115,6 +115,23 @@ When the key specified is missing or resolves to an invalid value, a correspondi
...
+- Whether the delay should occur before or after the dynamic timestamp
+- The `key` for the dynamic timestamp (which can come from your trigger data, recipient data, or other sources)
+
+As in the dynamic duration section above, the key specified must be available to be resolved. If the key is missing or resolves to an invalid value, a corresponding error will be logged on the workflow run, and the delay will be **skipped**.
```
samseely, line 128:
```suggestion
As in the dynamic duration section above, the key specified must be available to be resolved. If the key is missing or resolves to an invalid value, a corresponding error will be logged on the workflow run, and the delay will be skipped.
```
don't think you need the extra emphasis here

content/designing-workflows/delay-function.mdx
```diff
@@ -115,6 +115,23 @@ When the key specified is missing or resolves to an invalid value, a correspondi
...
+
+Relative interval delay functions are helpful for various scenarios, including:
+
+- Appointment reminders: Send a notification 1 day before an appointment time
```
samseely, line 132:
```suggestion
- Appointment reminders: send a notification one day before an appointment time
```

content/designing-workflows/delay-function.mdx
```diff
@@ -115,6 +115,23 @@ When the key specified is missing or resolves to an invalid value, a correspondi
...
+- Follow-up reminders: Send a follow-up message 2 hours after an event
```
samseely, line 133:
```suggestion
- Follow-up reminders: send a follow-up message two hours after an event
```

samseely (commented):
A few general notes outside of changed lines: 
* Up top on this file we call out: There are two types of delays we support in Knock today: "wait for set interval" and "wait for a dynamic duration."— we should add relative to this
* We should change any references throughout this page from "set interval" to "fixed interval" since that's how it's labeled in the UI.

natperpepaj:
> A few general notes outside of changed lines:
> 
> * Up top on this file we call out: There are two types of delays we support in Knock today: "wait for set interval" and "wait for a dynamic duration."— we should add relative to this
> * We should change any references throughout this page from "set interval" to "fixed interval" since that's how it's labeled in the UI.

going to move this back to draft - seems like there are a few changes even to the existing delay step types we can update so will reopen when ready

content/designing-workflows/delay-function.mdx
natperpepaj: @cellomatt I removed references to dynamic duration where we aren't explicitly referencing a duration unit format

content/designing-workflows/delay-function.mdx
```diff
@@ -115,6 +115,23 @@ When the key specified is missing or resolves to an invalid value, a correspondi
...
+- A fixed delay interval (provided by you in the workflow editor)
+- Whether the delay should occur before or after the dynamic timestamp
+- The `key` for the dynamic timestamp (which can come from your trigger data, recipient data, or other sources)
+
+As in the dynamic delay section above, the key specified must be available to be resolved. If the key is missing or resolves to an invalid value, a corresponding error will be logged on the workflow run, and the delay will be skipped.
+
+Relative interval delay functions are helpful for various scenarios, including:
+
+- Appointment reminders: send a notification one day before an appointment time
+- Follow-up reminders: send a follow-up message two hours after an event
```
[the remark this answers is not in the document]
natperpepaj: sounds good - I'll probably remove the ending sentence at the end of the first paragraph. maybe add a bit more explanation instead

PR #643, opened 2024-10-11
chore: restructure slack docs

JEverhart383:
### Description

This PR restructures all of the Slack-related docs found in 'Integrations' section to address the following issues in the current docs:

1. We offer the user many alternative paths, Slack, SlackKit, DIY, without making it clear which path they should take based on their use case. We have an opportunity to be more prescriptive here. 
2. This section of the docs combines reference material with implementation instructions. We should start with a clear explanation of what a Slack <-> Knock integration looks like, which is basically some combination of channel data.


samseely (approved):
💯

PR #655, opened 2024-10-29
docs: mention version of React Native used to test SDK

mattmikolay:
### Description

We previously received feedback that our docs don’t make it clear which versions of React Native our SDK can be used with. While we don’t currently test the SDK with multiple versions of React Native, we can at least mention that we test with React Native v0.73.4 internally.

### Screenshots

Screenshot 2024-10-29 at 9 57 59 AM


rachael-t (approved):
Thank you for adding this! 🙌

PR #657, opened 2024-10-30
chore: add callouts for inline identify with source events

cellomatt:
### Description

Due to customer confusion, this PR explicitly calls out that the test runner will only work with existing users for source events (meaning inline identification is not supported). It also updates some outdated links and adds an additional link to the documentation of what inline identification is.

When you use the rest runner (which triggers a workflow run for the most-recently-saved version of a workflow rather than hitting our public API), we don't support inline identification. For API-triggered workflows, the test runner modal allows you to select from a list of existing users. However, a source event-triggered workflow allows you to simply paste in a test event (with any recipient(s) that you would like to test with), meaning that it's possible to use inline identification syntax in your event.

If you do this, we'll show an error that the recipient is "missing" in the current environment, and we won't execute a workflow run. These callouts are an attempt to help a customer anticipate these errors before they occur and provide an alternative for testing inline identification via source events.

https://docs-git-mk-inline-test-runner-update-knocklabs.vercel.app/integrations/sources/overview#inline-identify-users-in-a-source-event

### DIFF
--- a/content/integrations/sources/overview.mdx
+++ b/content/integrations/sources/overview.mdx
@@ -113,7 +113,22 @@ Note that Knock will correctly map `name`, `email`, `avatar`, and `phone` proper
 
 ### Inline identify users in a source event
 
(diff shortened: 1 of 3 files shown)


rachael-t (approved):
Love it! 🎃

meryldakin (approved):
makes sense, thanks matt!

PR #663, opened 2024-11-13
fix: delete stub that was indexed into Algolia

cellomatt:
### Description

Remove Preferences quick start stub that was indexed into search


JEverhart383 (approved):
thanks for catching @woodke

PR #665, opened 2024-11-14
fix: update SupportModal to trigger Knock workflow

rachael-t:
### Description

Our SupportModal feedback was still being sent to Airtable, which we are no longer using to capture feedback. This update is to instead have the submission trigger a Knock workflow to send an email to the DSE team to review and action any feedback related to the docs. 

### Tasks

[KNO-7382](https://linear.app/knock/issue/KNO-7382/[docs]-update-supportmodal-to-not-use-airtable)


cellomatt (approved):
10/10 🚀 thank you!

PR #666, opened 2024-11-14
feat: add docs for batch w/ leading debounce

cjbell:
### Description

Adding docs for batch w/ leading debounce. Very open to feedback!

### Tasks

KNO-7305


cellomatt (approved):
LGTM! I think we should also cut lines 396-398 in the FAQs 😀

PR #673, opened 2024-11-19
chore: add requirement indicators for workflow trigger endpoints

rachael-t:
### Description

Currently, this PR adds missing requirement indicators to clarify whether API attributes/parameters are required or optional for the workflow trigger endpoints _only_. Looking throughout the entire API reference, we inconsistently document this, and I have filed [this Linear ticket](https://linear.app/knock/issue/KNO-7408/[docs]-add-missing-requirements-indicators-to-api-attributesparameters) to capture the larger update that is needed.

### DIFF
--- a/content/reference.mdx
+++ b/content/reference.mdx
@@ -571,7 +571,7 @@ The workflow trigger endpoint enforces the following payload size limits:
 <Attributes>
   <Attribute
     name="recipients"
-    type="RecipientIdentifier[]"
+    type="RecipientIdentifier[] (required)"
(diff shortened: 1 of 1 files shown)


cellomatt:
re: the broader inconsistency that you called out, I almost wonder if we should consider adding an `isRequired` prop to the `<Attribute>` component that displays some sort of `required` or `optional` tag, it feels a bit weird to me that we're putting this into the `type` field right now

cellomatt (approved):
lgtm!

PR #688, opened 2024-11-26
chore: update courier migration guide

cellomatt:
### Description

Update Courier migration guide in light of some new documentation/product updates they've made. I also added a mention of Partials, plus additional mAPI endpoints for Partials and Layouts.

Note: some info is missing [here](https://www.courier.com/docs/reference/notifications/get-notifications), but I updated to that link because the old link (https://www.courier.com/docs/reference/notifications/list) is broken. It seems like they may be adding templates to their API, at which point I can update the section on migrating workflows/notification templates.


samseely (approved):
Thank you for remembering to update this!

PR #689, opened 2024-11-26
feat: add docs about overriding the to address for emails

cjbell:
### Description

Adds docs for overriding the `to` address for all emails

### Tasks

KNO-7458

### DIFF
--- a/content/integrations/email/settings.mdx
+++ b/content/integrations/email/settings.mdx
@@ -6,7 +6,7 @@ layout: integrations
 tags: ["bcc", "cc", "JSON overrides", "email settings", "email overrides"]
 ---
 
-Knock email channel configurations support a number of settings. These include include email-specific fields (such as cc, bcc, and reply-to) as well as JSON overrides to be pass in API calls to the configured provider's API endpoints.
+Knock email channel configurations support a number of settings. These include email-specific fields (such as `cc`, `bcc`, and `reply-to`) as well as JSON overrides to be passed in API calls to the configured provider's API endpoints.
 
 When you configure a setting in a channel's configuration, it will be used on all instances of that channel across all workflows.
 
@@ -14,8 +14,8 @@ When you configure a setting in a channel's configuration, it will be used on al
 
 You can override email settings on a per-channel basis, or on a per-template basis.
 
-- **At the channel-level**. The to, cc, bcc, and reply-to fields can be found in the "Overrides" section of the channel's "Settings" tab.
-- **At the template-level**. You'll find the email settings under the "Manage template settings" button.
+- **At the channel level**. The `to`, `cc`, `bcc`, and `reply-to` fields can be found in the "Overrides" section of the channel's "Settings" tab.
+- **At the template level**. You'll find the email settings under the "Manage template settings" button.
 
 All email configuration fields support Liquid usage. You will be able to use any variables available in your workflow trigger payloads, as well as system variables such as the current workflow, activities, and any other variables you have access to when building templates.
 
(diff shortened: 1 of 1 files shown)


content/integrations/email/settings.mdx
```diff
@@ -1,23 +1,21 @@
...
 layout: integrations
-tags: ["bcc", "cc", "JSON overrides"]
+tags: ["bcc", "cc", "JSON overrides", "email settings", "email overrides"]
 ---
 
-In this guide we cover how to configure email channels in Knock.
-
-## An intro to email channel settings
-
 Knock email channel configurations support a number of settings. These include include email-specific fields (such as cc, bcc, and reply-to) as well as JSON overrides to be pass in API calls to the configured provider's API endpoints.
```
cellomatt, line 9:
```suggestion
Knock email channel configurations support a number of settings. These include email-specific fields (such as `cc`, `bcc`, and `reply-to`) as well as JSON overrides to be passed in API calls to the configured provider's API endpoints.
```
Not your change, but these should be fixed 😅 

Not sure about `code` style on these fields but if we add here we should be consistent below as well

content/integrations/email/settings.mdx
```diff
@@ -1,23 +1,21 @@
...
+When you configure a setting in a channel's configuration, it will be used on all instances of that channel across all workflows.
 
-### Configuring email settings
+## Configuring email settings
 
-The cc, bcc, and reply-to fields can be found at the bottom of the channel environment configuration modal's "Settings" tab.
-The JSON overrides field can be found under the "Overrides" Tab.
+You can override email settings on a per-channel basis, or on a per-template basis.
+
+- **At the channel-level**. The to, cc, bcc, and reply-to fields can be found in the "Overrides" section of the channel's "Settings" tab.
```
cellomatt, line 17:
```suggestion
- **At the channel level**. The `to`, `cc`, `bcc`, and `reply-to` fields can be found in the "Overrides" section of the channel's "Settings" tab.
```
nit, don't think hyphen is necessary. also added `code` style to fields

content/integrations/email/settings.mdx
```diff
@@ -1,23 +1,21 @@
...
+- **At the template-level**. You'll find the email settings under the "Manage template settings" button.
```
cellomatt, line 18:
```suggestion
- **At the template level**. You'll find the email settings under the "Manage template settings" button.
```
nit

meryldakin (approved):

content/integrations/email/settings.mdx
```diff
@@ -27,7 +25,13 @@ As an example, if you wanted to conditionally change the "From name" on an email
 {{ actor.from_name | default: "no-reply@knock.app" }}
 ```
 
-### CC and BCC fields
+## Overriding the default `to` address
+
+By default Knock will send your emails to the `email` property stored on the `recipient` for the workflow run. If you need to override this, you can do so by setting the `to` field in your email configuration either at the channel or the template level.
+
+As an example, if you wanted to send all emails to a single address, you could set the `to` field to that address at the channel to either a static value (like `hello@example.com`) or a dynamic value (like `{{ data.email_to_override }}`).
```
cellomatt, line 32:
```suggestion
As an example, if you wanted to send all emails to a single address, you could set the `to` field at the channel level to either a static value (like `hello@example.com`) or a dynamic value (like `{{ data.email_to_override }}`).
```

content/integrations/email/settings.mdx
```diff
@@ -27,7 +25,13 @@ As an example, if you wanted to conditionally change the "From name" on an email
...
+
+## Setting `cc` and `bcc` addresses
 
 The cc and bcc fields can be used to support a single or multiple addresses. In order to use several addresses on any of these fields, make sure to separate them with a comma.
```
cellomatt, line 36:
```suggestion
The `cc` and `bcc` fields can be used to support a single or multiple addresses. In order to use several addresses on any of these fields, make sure to separate them with a comma.
```
Just for consistency

cellomatt (approved):
Left a few suggestions

PR #697, opened 2024-12-07
chore: add guidance on using knock as template source of truth

cellomatt:
### Description

This PR [adds an FAQ](https://docs-8z71sj3av-knocklabs.vercel.app/designing-workflows/template-editor/overview#frequently-asked-questions) to the docs on **Why store all of my notification templates in Knock?**

### Screenshots

image

### DIFF
--- a/content/designing-workflows/template-editor/overview.mdx
+++ b/content/designing-workflows/template-editor/overview.mdx
@@ -183,32 +183,22 @@ To test your notification, click "Save" and navigate back to the workflow canvas
 
 All test runner notifications are tracked under "Messages" and have a source of "Test runner" to distinguish them from notifications you've triggered via the Knock API.
 
-## Why store all of my notification templates in Knock?
+## Frequently asked questions
 
-This is a common question that we hear from many of our customers, particularly those who currently maintain templates for their notifications in other third-party tools (often more than one!).
+<AccordionGroup>
+  <Accordion title="Why store all of my notification templates in Knock?">
+    This is a common question that we hear from our customers, particularly those who currently maintain templates for their notifications in other third-party tools (often more than one!).
 
-While [planning a migration](/guides/implementation-guide) will require some upfront effort, we recommend using Knock as the source of truth for your notification templates for a number of reasons:
+    While [planning a migration](/guides/implementation-guide) will require some upfront effort, we recommend using Knock as the source of truth for your notification templates for a number of reasons:
 
-<Steps>
-  <Step title="A consistent editing experience">
-    When you manage all of your templates in Knock, your content authors have access to a unified template editor for a consistent experience across all delivery channels. Teams who collaborate on messaging can work with a single templating language (Liquid) for dynamic content.
-  </Step>
-  <Step title="Cross-channel engagement analytics">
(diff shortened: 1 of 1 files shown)


content/designing-workflows/template-editor/overview.mdx
```diff
@@ -181,3 +182,33 @@ Your notification preview is populated with the data available in the lefthand v
...
+  </Step>
+  <Step title="Work with your templates programmatically">
+    With our [Management API](/mapi) or [CLI](/cli), you can work with all of your cross-channel templates as code via a single API. This enables you to integrate your templates with other tools, automate updates, and more.
+  </Step>
+  <Step title="Reusable content blocks">
+    Our [Partials](/designing-workflows/partials) feature allows you to create content blocks of various types that can be reused across all of your notification templates. This enables updates across all delivery channels with a single change when necessary.
+  </Step>
+  <Step title="Reference data from other resources stored in Knock">
+    In addition to the context that you pass to Knock on your workflow trigger calls and [reference as dynamic content](/designing-workflows/template-editor/variables) in your notifications (like recipient and `tenant` properties, or custom `data` payload variables), the Knock template editor also allows you to [reference data](/designing-workflows/template-editor/referencing-data) from any Users, Objects, and Tenants that exist within your Knock environment.
```
rachael-t, line 209:
```suggestion
    In addition to the context that you pass to Knock on your workflow trigger calls and [reference as dynamic content](/designing-workflows/template-editor/variables) in your notifications (like `recipient` and `tenant` properties, or custom `data` payload variables), the Knock template editor also allows you to [reference data](/designing-workflows/template-editor/referencing-data) from any Users, Objects, and Tenants that exist within your Knock environment.
```

rachael-t:
My thoughts:

1. I feel as though this is the best place for this. I know we chatted about how it might also make sense as a blog post, but having it on this page feels like the best fit to me.
2. We also chatted about this one, and I think the `<Step>` component looks best. If we were to do just a `h4` my vote would be to do so without the numbering. 
3. At first glance, it read well to me! I personally like 6 being separated and love that the section ends with "and isn't possible with other templating solutions," as that felt more compelling to me.
4. I think so!

samseely:
I like this addition a lot. Reading it, it feels like this is the right page for the content but the H2 feels a bit awkward to me alongside the rest of the table of contents of this page. 

Can we try it in an FAQ? If it is a common question we hear from customers, then that feels like the right place to put it since it's a pattern we use on other pages in the docs (and one we should probably remember to add to docs like Audiences concept doc and keep updated as we get new questions). 

If we do put it under an FAQ, it will be a longer one, but I think that's okay and we can take it out of <Step/> and just use an unordered list with bold intros at the start of each one. 

E.g. 

* **A consistent editing experience.** When you manage all your...
* **Cross-channel engagement analytics.** With Knock link and...

But yes, 100% compelling and an item we should callout to Hashim and sales team so they know it exists and can link to it from [sales enablement assets DB](https://www.notion.so/knockapp/Approved-sales-assets-128c5a209978803cb8a3c449f47fc111) in notion

samseely (approved):
Approving so you can merge once feedback is incorporated. My main point was making this a part of a broader FAQ section that we can add other common questions to but let me know if you have other questions on my notes.

rachael-t (approved):
I like it as an FAQ! Great change

PR #707, opened 2024-12-12
Update overview.mdx

sbcooney:
iOS page accidentally referenced React instead of iOS.


rachael-t (approved):
Thank you for flagging this! 🙏

PR #708, opened 2024-12-12
feat: introduce isRequired for Attribute component

rachael-t:
### Description

Introduces the `isRequired`prop to the `<Attribute>` component, which, when provided, displays a "required" tag.

This PR is _only_ to introduce the prop. I will open a seperate PR with the updates to the API + mAPI docs. 

### Tasks

[KNO-7408](https://linear.app/knock/issue/KNO-7408/[docs]-add-missing-requirements-indicators-to-api-attributesparameters)

### Screenshots

Current:
Screenshot 2024-12-12 at 12 54 45 PM

With new `isRequired` prop:
Screenshot 2024-12-12 at 12 49 47 PM

Current:
Screenshot 2024-12-12 at 12 55 30 PM

With new `isRequired` prop:
Screenshot 2024-12-12 at 12 47 53 PM

### DIFF
--- a/components/Attributes.tsx
+++ b/components/Attributes.tsx
@@ -40,7 +40,7 @@ const Attribute = ({
         )}
       </span>
       {isRequired && (
-        <span className="font-semibold text-brand bg-beige-light text-xs ml-2 py-0.5 px-1">
+        <span className="font-semibold text-brand dark:text-brand-light bg-beige-light dark:bg-zinc-800 text-xs ml-2 py-0.5 px-1">
           required
         </span>
       )}
--- a/components/Attributes.tsx
+++ b/components/Attributes.tsx
@@ -40,9 +40,15 @@ const Attribute = ({
         )}
       </span>
       {isRequired && (
-        <span className="font-semibold text-brand dark:text-brand-light bg-beige-light dark:bg-zinc-800 text-xs ml-2 py-0.5 px-1">
-          required
-        </span>
+        // option 1
+        <span className="font-semibold text-brand-dark bg-brand-light/20 dark:text-brand-light dark:bg-brand-dark/20 text-xs ml-2 py-0.5 px-1">
+  required
+</span>
+        // option 2
+//         <span className="font-semibold text-brand-dark bg-brand-light/10 dark:text-brand-light dark:bg-zinc-800/30 text-xs ml-2 py-0.5 px-1">
+//   required
+// </span>
+
       )}
     </span>
     <span className="block text-sm mt-0 text-gray-600 dark:text-gray-300">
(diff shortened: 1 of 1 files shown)


cellomatt:
Love it!

A suggestion:
- I think we need `dark` styling for this as well

A couple open questions:
- Should we use a different color for this? The red brand color shows on the linked `Types` and I think it's nice that it indicates they are clickable
- I am wondering if we would also want to extend this to have three options: default to blank, explicitly `required`, or explicitly `optional`? Maybe that's unnecessary to have all three, but it's worth considering the pattern of "required" vs. "optional" and which way we might want to go there

rachael-t:
> * I think we need `dark` styling for this as well
Ah yup - had it in there then did some changes and apparently dropped that. Just pushed up some changes with dark mode added:
Screenshot 2024-12-13 at 8 48 10 AM

> * Should we use a different color for this? The red brand color shows on the linked `Types` and I think it's nice that it indicates they are clickable
I went back and forth on this. I like red because it calls out that it's important (i.e., this is required, so make sure to add it), but I agree that if there is a `typeSlug` preceding it, it can be a lot of red. 

> * I am wondering if we would also want to extend this to have three options: default to blank, explicitly `required`, or explicitly `optional`? Maybe that's unnecessary to have all three, but it's worth considering the pattern of "required" vs. "optional" and which way we might want to go there
I'd be interested to hear what others think about this as well! Right now, our usage is inconsistent. We usually call out when something is required, but we sometimes use optional and sometimes do not (and I think that's confusing). I personally think just calling out when something is required will help it stand out more (versus there being too many unnecessary tags), and then the optional is implied. Open to having a discussion on this!

cellomatt:
@rachael-t I like that idea of the background color but not sure about red-on-red... I'm wondering if we should ask for Krisna's input on this? The problem I'm hoping to solve here is 
- differentiate the clickable `Type` link from this label
- still make this label clear and obvious

> I personally think just calling out when something is required will help it stand out more (versus there being too many unnecessary tags), and then the optional is implied.

I like this, that makes sense to me, and obviously a "later on" thing to be added to the style guide and updated once this PR is taken care of

rachael-t:
Updated using Option 2 per the Slack conversation. I also added a border for dark mode to align with other tags. 

Here is light mode:
Screenshot 2024-12-26 at 10 15 02 AM

Here is dark mode:
Screenshot 2024-12-26 at 10 15 09 AM

Note that [I checked](https://knocklabs.slack.com/archives/C068SUY4C06/p1734645282752769?thread_ts=1734624148.361559&cid=C068SUY4C06) to ensure that both light and dark modes were WCAG AA compliant.

mattmikolay (approved):
🚀 🚀 🚀

PR #714, opened 2024-12-16
chore: add OneSignal to supported push channels

rachael-t:
### Description

OneSignal was missing from the list of push channels supported on the [What is Knock?](https://docs.knock.app/getting-started/what-is-knock#channels) page, and this PR is to add it in: https://docs-git-rt-add-onesignal-to-channels-knocklabs.vercel.app/getting-started/what-is-knock#channels


cellomatt (approved):
Good catch.

Not sure if you want to add it to this PR, but it looks like we're also missing [Sinch MessageMedia](https://docs.knock.app/integrations/sms/sinch-message-media). 

I also wonder if we should remove the accordion altogether here and link back to the Integrations overview? I know it's not as cute, but might be better than having the possibility of forgetting to maintain this. Alternatively, we could add a comment to that overview page to remind folks to update the accordion here if they are also updating the list there 🤷

PR #719, opened 2024-12-19
chore: add docs for twilio bounce support

cjbell:
### Description

Callout that twilio supports bounces


cellomatt (approved):
💯 lgtm!

Makes me realize we should do an audit of our integration provider pages, we don't document **Features** consistently right now.

PR #723, opened 2024-12-23
fix(kno-7538): update images to use Image component

rachael-t:
### Description

Many of the images used throughout the docs did not follow our [style guide](https://www.notion.so/knockapp/Knock-writing-style-guide-9fd5eaf44be54994ad8741d1c86f20f1?pvs=4#1bf025dd17bf42668b54fe4115f63862) and use the `<Image>` component (with the classes `className="rounded-md mx-auto border border-gray-200"`)

This PR updates those image references to use the component. In the cases where an image was either very outdated, or either had a border in the image itself, I created and added a new image. 

Note that I did **not** update `.gif` references to use the `<Image>` [component](https://nextjs.org/docs/pages/api-reference/components/image) since the component is designed for static image optimization.

From this work, a few other tasks have been flagged in the Docs project in Linear:
1. Update images with template editor v3 release
2. Add a checklist to `pull_request_template.md`
3. Add CI/build rules to enforce image formatting

### Todos

- [x] [Update images that were flow charts or tables](https://linear.app/knock/issue/KNO-7538/[docs]-update-images-so-all-are-using-the-image-component#comment-2040cfa5) - will not use the `border border-gray-200` classes
  - [x] Confirm with team that's how we want to handle these types of images, and if so, update style guide
- [x] [Update images that are of a full browser window](https://linear.app/knock/issue/KNO-7538/[docs]-update-images-so-all-are-using-the-image-component#comment-b4b887fd) (and therefore have formatting applied) - will not add any classes
  - [x] Confirm with team that's how we want to handle these types of images, and if so, update style guide
- [x] Triple check that any deleted images are not being used anywhere in the docs

### Tasks

[KNO-7538](https://linear.app/knock/issue/KNO-7538/[docs]-update-images-so-all-are-using-the-image-component)


content/guides/customer-webhooks.mdx
```diff
@@ -99,15 +99,14 @@ If you are passing a JSON payload to the webhook, you can use the `json` filter
 
 Now we can build our notification workflow that sends out a webhook when the customer has a webhook configured for the object.
 
-Start by creating a new workflow in Knock:
```
rachael-t, line 102: Regarding the changes below, see [KNO-7573](https://linear.app/knock/issue/KNO-7573/[docs]-update-customer-webhook-guide)

rachael-t:
I'd like the team to weigh in on the following decisions I made for certain image types. Based on what we decide, I will update the style guide.

1. Images that are flow charts/tables should not use the `border border-gray-200` classes
2. images that are of a full browser window (and therefore have formatting applied) will not have any classes applied

rachael-t:
I realized that the [Fetch function page](https://docs-git-rt-kno-7538-update-images-knocklabs.vercel.app/designing-workflows/fetch-function) has double borders (and the UI in the images is outdated). However, I will need to make that update once I have access to a larger monitor for better screenshots ([Loom of me explaining](https://www.loom.com/share/59cb8fa753654857adb29e79ebb2bd12?sid=8dc7a503-1230-4546-8a26-b02c09648218)).

content/in-app-ui/overview.mdx
```diff
@@ -18,7 +24,13 @@ Knock differs from other customer messaging platforms by separating the **conten
 
 ## How Knock powers in-app messages
 
-![How Knock powers in-app messages](/images/in-app-ui/how-knock-powers-in-app.png)
+<Image
+  src="/images/in-app-ui/how-knock-powers-in-app.png"
+  width={800}
+  height={450}
+  className="rounded-md mx-auto"
```
cellomatt, line 31: nit: can probably cut this line since we aren't applying the border (these other styles aren't really relevant since we can't see them)

content/integrations/extensions/segment.mdx
```diff
@@ -38,9 +38,11 @@ Knock uses the Segment HTTP Source to send events to your Segment account. Each
...
 5. Once you click save, events will begin flowing from that environment to Segment
 
-<img
-  width="400"
-  src="/images/integrations/extensions/segment_modal.png"
+<Image
+  src="/images/integrations/extensions/connect-segment-modal.png"
+  width={800}
+  height={450}
+  className="rounded-md mx-auto border border-gray-200"
```
cellomatt, line 45: One thing I wonder about is whether we might want to add a constrained width to some of these images (esp. relevant here because the original had that). 

Supposedly the height/width props on the `Image` component are not used to style the image, but I tried setting `width={400}` on this one locally and it did constrain it. Sometimes we have images like this one that look kind of oversized when rendered to the full column width.

content/send-notifications/message-statuses.mdx
```diff
@@ -18,7 +18,13 @@ When you trigger a Knock workflow, any channel step within that workflow generat
 
 You can use the "Logs" tab in the message detail view in the Knock dashboard to examine the history of requests Knock makes to the downstream provider to determine delivery status. The "Delivery status" field will show you the current status for your message.
 
-![Knock message delivery status lifecycle](/images/notifications/delivery-status-lifecycle.png)
+<Image
+  src="/images/notifications/delivery-status-lifecycle.png"
+  width={800}
+  height={450}
+  className="rounded-md mx-auto"
```
cellomatt, line 25: nit: same as above, should cut

cellomatt (approved):
Marking as approved so you can move forward, I confirmed that all of the images (including renamed ones) are displaying properly.

With that being said -- I think we should review the usage of `width`/`height` across all of these images, and also come up with some guidelines for their use to add to the style guide. It's unclear to me how exactly they are used by Next, as it says that they are not CSS styles but in my own testing, they do seem to impact the way that images are rendered.

I think these values should probably match the dimensions of the original image, or at least fractionally (like `400` for an image with 800px width, for example). I also don't love the way that a lot of these images look when they render 100% width on the content column, and think it could be worth considering when/if to constrain the width or height of certain images

content/in-app-ui/overview.mdx
rachael-t: I thought the same, then realized they were visible in dark mode. Wanted to make sure that even in dark mode, the images all have a consistent look.

cellomatt: Oh, that's a great call -- did not even think of that. Good catch!!

rachael-t:
@cellomatt, I think this is finally in a good place! **Not urgent** but when you have a moment, can you give things another once-over? I've reviewed every page in the [preview link](https://docs-git-rt-kno-7538-update-images-knocklabs.vercel.app/) a few more times, and all images are rendering and looking correct. I just would like another gut check whenever you can since it's a big update. 😅

Also, I updated the [style guide](https://www.notion.so/knockapp/Knock-writing-style-guide-9fd5eaf44be54994ad8741d1c86f20f1?pvs=4#f0403ab0d6974bf78fd35189681da96b) with some best practice based on what was uncovered during this work.

content/concepts/audiences.mdx
```diff
@@ -39,7 +39,7 @@ When you create an audience, its key instantly exists across all environments. A
   height={926}
   width={470}
   alt="Audience selector in the workflow editor"
-  className="rounded-md border border-gray-200"
+  className="rounded-md mx-auto border border-gray-200"
```
cellomatt, line 42: Not actually related to a change you made, but I noticed the image on ln 80 of this file could probably be a good one to reduce the width on while we're at it

cellomatt (approved):
I happened to see one unrelated image that might be nice to update, but this is g2g as far as I'm concerned. thank you for doing this!!!! it's not glamorous but the consistency is amazing and it fixes so many weird formatting things 👏

content/concepts/audiences.mdx
cellomatt: (I couldn't comment in the right place because it was unrelated)

rachael-t: Oh good catch!

PR #727, opened 2025-01-03
chore: "on this page" sidebar improvements (KNO-7487)

cellomatt:
### Description

This PR includes a few improvements to the "On this page" `<aside>` element that renders a sidebar of the headings on the current page above certain screen widths:

- Increases the width of the sidebar slightly so the instances where we need a horizontal scroll are less likely to occur
- Adds some padding-right to the listed headers so that _when_ horizontal overflow occurs, there is some space on the right side when scrolled all the way to the right
- Slight adjustment to height of the sidebar to make it slightly taller
- Adds a gradient to the bottom of the sidebar so that you can visually tell when there is overflowed content to scroll (in both light and dark mode)
- Wraps the inner content in an additional div so that bottom padding doesn't make the gradient appear above the padding

### Notes

I went back and forth on how fancy to get with this -- we could try to do something that renders conditionally based on whether or not the content overflows the container, for example -- but for simplicity I opted to just go the route of always displaying the gradient and you'll only see it if the content reaches the full container height.

This meant also making a decision about CSS positioning, and whether the gradient div "takes up space" or is a simple overlay. Here I decided to have it take up space at the bottom of the container, because otherwise scrolling all the way to the bottom (or, in instances where there is no overflow but the content takes up the full height of the element) would mean that some of the content at the bottom would always be behind the gradient.

These are definitely compromises, but I wasn't sure if this was worth totally re-working in the short term.

### Screenshots

image
image

### DIFF
--- a/components/PageNav.tsx
+++ b/components/PageNav.tsx
@@ -122,7 +122,10 @@ const PageNav: React.FC<Props> = ({ title, sourcePath }) => {
         >
           Edit this page on GitHub &rarr;
         </a>
-        <div id="gradient" className="sticky bottom-0 left-0 right-0 h-24 pointer-events-none bg-gradient-to-t from-white dark:from-gray-900 to-transparent" />
+        <div
+          id="gradient"
+          className="sticky bottom-0 left-0 right-0 h-24 pointer-events-none bg-gradient-to-t from-white dark:from-gray-900 to-transparent"
+        />
       </div>
     </aside>
   );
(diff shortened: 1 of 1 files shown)


rachael-t (approved):
I really love the gradient 🤩

samseely (approved):
💯 — love this! I'll defer to Chris to any of the questions you had on CSS positioning etc but what you have on the preview link looks great to me

cjbell (approved):
Very happy with approach. I would have done the same thing.

PR #732, opened 2025-01-09
feat(KNO-7524): update docs to add MsTeamsAuthButton

mattmikolay:
### Description

This PR updates the documentation to reflect the addition of `MsTeamsAuthButton`, the first component of TeamsKit. I used the existing SlackKit documentation as reference.

### Tasks

[KNO-7524](https://linear.app/knock/issue/KNO-7524/[docs]-add-docs-for-msteamsauthbutton)


content/managing-recipients/setting-channel-data.mdx
```diff
@@ -215,15 +215,15 @@ Channel data requirements for each provider are listed below. Typically `channel
 
     | Property            | Type     | Description                    |
     | ------------------- | -------- | ------------------------------ |
-    | ms_teams_tenant_id  | `string` | A Microsoft Teams tenant ID    |
+    | ms_teams_tenant_id  | `string` | A Microsoft Entra tenant ID    |
```
mattmikolay, line 218: “Microsoft Entra tenant” is more accurate than “Microsoft Teams tenant”.

content/in-app-ui/react/teams-kit.mdx
```diff
@@ -0,0 +1,340 @@
...
+  <Step title="Add API permissions">
+    In the top-level sidebar, navigate to **API permissions** and click **Add a permission**. Select **Microsoft Graph** > **Application permissions**.
+
+    Under **Select permissions**, add the following permissions:
+
+    - `Team.ReadBasic.All`: This permission allows your bot to read a list of all teams within a Microsoft Entra tenant.
+    - `Channel.ReadBasic.All`: This permission allows your bot to read a list of all channels within a team.
+    - `AppCatalog.Read.All`: This permission allows your bot to locate itself in the Microsoft Teams app catalog.
+    - `TeamsAppInstallation.ReadWriteSelfForTeam.All`: This permission allows your bot to install itself into a team.
+    - `TeamsAppInstallation.ReadWriteSelfForUser.All`: This permission allows your bot to install itself into a user's personal scope.
```
mattmikolay, line 50: I started [a thread in Slack](https://knocklabs.slack.com/archives/C07TREA8R45/p1736455798947329) to discuss these permissions.

content/in-app-ui/react/teams-kit.mdx
```diff
@@ -0,0 +1,340 @@
...
+## Getting started
+
+To get started you'll need a [Knock account](https://dashboard.knock.app), a [Microsoft Teams channel connected to a Microsoft Teams bot](/integrations/chat/microsoft-teams/overview), and a workflow with a Microsoft Teams channel step.
+
+Depending on your use case, you can follow one of these guides for step-by-step instructions on how to set up your Microsoft Teams integration:
+
+- [Sending messages to public channels](/integrations/chat/microsoft-teams/sending-a-message-to-channels) in your customers' Microsoft Teams instances
+- [Sending direct messages](/integrations/chat/microsoft-teams/sending-a-direct-message) to users in your customers' Microsoft Teams instances
+
+## Configuring your bot in Microsoft Entra
```
JEverhart383, line 18: I feel like this section on configuring the Bot might be better suited for the integrations overview page (/integrations/chat/microsoft-teams/overview), at least that's how I tried to structure the SlackKit docs. IMO, this page works best if we just focus on the frontend components, not also the backend integration requirements.

content/integrations/chat/microsoft-teams/sending-a-direct-message.mdx
```diff
@@ -69,22 +72,122 @@ In your `manifest.json` file, add `personal` to your bot's array of scopes:
...
+        target="_blank"
+      >
+        preinstall your bot for all Microsoft Teams users in their organization
+      </a>
+      .
+    </>
+  }
+/>
+
+## Setting User channel data
```
JEverhart383, line 190: If I understand this code sample correctly, the `onInstallationUpdateAdd` callback will fire every time the bot is installed, and I can see where we extract the `msTeamsUserId` from the activity. But is there anything else contained in this payload that could help the developer map this teams user id to a user in Knock?

I'm imagining if there were an `email` field, I'd query my own DB to get a user and their id, then use that ID to set channel data.  Would I need to use another MS Teams API to resolve the Ms teams user id to an email? Looking at the linked docs didn't help me much, as I couldn't find a payload definition. Ideally what we'd provide here would be as close to drop-in as possible.

Does this also mean that someone would be required to use the bot framework to enable DMs with Teams?

JEverhart383: Ahh I see now the answer to my last question: https://dev.botframework.com/
Seems like this is the equivalent of Slack's bolt framework, but is required to use a MS teams bot.

content/in-app-ui/react/teams-kit.mdx
JEverhart383: Just to explain a little bit of rationale here as well: the first iteration of the SlackKit docs had some instructions on completing the "integration" on the component pages, but obviously they weren't complete. Because there was enough text there, people wouldn't look beyond the component page for instructions, get confused when things didn't work, and reach out for support.

Once we got to the third iteration of SlackKit docs, I tried to break things up like this so each doc had a specific purpose:
- Overview: this was all the necessary backend config someone would need to complete and conceptual information about how channel data fit into this
- Use case guides: a specific and prescriptive how-to on implementing a particular use case or pattern
- Component pages: a slightly more detailed reference for each component and what a frontend developer might need to implement 

Happy to talk through this pattern if we think it might not make as much sense for TeamsKit. I have nothing but anecdata to suggest that after our later revision, we've gotten fewer questions and those we do get tend have gotten further before reaching out.

content/integrations/chat/microsoft-teams/sending-a-direct-message.mdx
```diff
@@ -69,22 +72,122 @@ In your `manifest.json` file, add `personal` to your bot's array of scopes:
...
+The only access you'll need to manage when using TeamsKit are grants for your users to interact with their [Tenant](/concepts/tenants) in Knock. This is necessary because the user in this context is an end user in your application who does not have access to Knock as a [member of the account](/manage-your-account/managing-members). Therefore, these grants provide them elevated privileges to operate on specific resources using the API.
 
-If you're using Knock's Node SDK, your code might look like this:
+We've made it easy for you to tell Knock which resources your users should have access to by making it a part of their user token. In this section you'll learn how to generate these grants using the Node SDK and, if you're not using the SDK, how to structure them for other languages.
+
+You'll need to generate a token for your user that includes access to the Knock tenant storing the Microsoft Entra tenant ID as well as any recipient objects storing Slack channel data described in this reference on [TeamsKit resource access grants](/in-app-ui/react/teams-kit#resource-access-grants).
```
meryldakin, line 85: I don't think we mean Slack here right?

content/integrations/chat/microsoft-teams/sending-a-direct-message.mdx
mattmikolay: So `msTeamsUserId` here is an ID that’s unique to both the Microsoft Teams user _and_ the bot. Also present under `from` is an `aadObjectId`, which is the unique Entra ID of the user. Using `aadObjectId`, the application could make a network request to the [get user endpoint of the Graph API](https://learn.microsoft.com/en-us/graph/api/user-get) to look up the email address for this user. Then, the application could query the DB to get a user ID for use with Knock, just as you described.

Microsoft does provide a [JS library for use with the Graph API](https://github.com/microsoftgraph/msgraph-sdk-javascript). I’ve never used it, but let me look into updating the code sample to make this request. If it’s too convoluted, I’ll at least update the text to describe what the application should do.

> Does this also mean that someone would be required to use the bot framework to enable DMs with Teams?

Yes, from what I understand, building a Microsoft Teams bot always requires the use of the Bot Framework. When we (Knock) send a message as a Microsoft Teams bot to either a Teams channel or a user’s personal scope (i.e. direct message), we use the [Bot Framework REST API](https://learn.microsoft.com/en-us/azure/bot-service/rest-api/bot-framework-rest-connector-api-reference).

content/integrations/chat/microsoft-teams/sending-a-direct-message.mdx
mattmikolay: Good catch! Fixed.

meryldakin (approved):
approving from my end so as not to block but I know there may still be some updates to the sections jeff was mentioning!

content/integrations/chat/microsoft-teams/sending-a-direct-message.mdx
JEverhart383: Gotcha, that's kind of different than how Slack's user id concept works, it's per-user and per-workspace, but not per-bot. If it's feasible to update the code sample, I'd recommend it. Shocked but not shocked that's not something that would be baked into the framework. But from our perspective, we'll be asked through support how to connect these dots, so if we can do it now it'll save us cycles down the road - we could also treat this as non-blocking as long as we annotate with the steps required and point them to the right API.

content/in-app-ui/react/teams-kit.mdx
mattmikolay: That makes sense! Thanks for explaining your rationale for organizing the SlackKit docs this way. I’ll move the Entra config instructions to the MS Teams overview page. 👍🏻

JEverhart383 (approved):
🔥 excited for this new feature

PR #750, opened 2025-01-16
chore(kno-7612): update all manage template settings references

rachael-t:
### Description

I noticed that in a few places in the docs, we still refer to the button in the template editor as "Manage template settings," but it is now simply "Template settings."

### Todos

- [x] Review images for references and update as needed

### Tasks

[KNO-7612](https://linear.app/knock/issue/KNO-7612/[docs]-update-manage-template-settings-references-to-template-settings)

### Screenshots

Current image:
Screenshot 2025-01-16 at 12 28 14 PM

New image:
Screenshot 2025-01-16 at 12 27 49 PM


cellomatt (approved):
Only q here is about the replacement image -- it seems like it's smaller than the other one and on my screen it looks sort of pixelated on the pages where it's rendered. Also not sure whether the height/width on the image component has any affect on that 🤔 
image

rachael-t:
@cellomatt I did want it to be slightly smaller, as I felt the original was a bit large. I did just adjust the width and height to be the original form the image and I believe that helped? Can you get you to double check? https://docs-git-rt-kno-7612-update-manage-template-se-15959a-knocklabs.vercel.app/sdks/ios/deep-links#2-include-a-deep-link-in-your-knock-message-payload

cellomatt:
> @cellomatt I did want it to be slightly smaller, as I felt the original was a bit large, but it did look a little off. I just adjusted the width and height to be the original form of the image, and I believe that helped. Can you get you to double-check? https://docs-git-rt-kno-7612-update-manage-template-se-15959a-knocklabs.vercel.app/sdks/ios/deep-links#2-include-a-deep-link-in-your-knock-message-payload

Hmm so it still looks blurry to me... I'm curious how you got this to be smaller, did you shrink your window size before taking the screenshot or did you reduce the size of a screenshot you had already taken?

I _think_ what might work best here is to set the "implicit" height/width on the `Image` component to be something like half the height and width of the actual file. So for example, a square image with a size of 1000x1000 would be set with width/height values of 500. This would effectively double the resolution but should still display the image at the smaller size in most scenarios (if I understand the Next.js documentation correctly)

PR #757, opened 2025-01-24
chore: update message delivery statuses

JEverhart383:
### Description

Slack community user pointed out that the `status` options in API reference don't match what's called out on message status docs. I cross-referenced with `MessageStatus` type in Node SDK and updated the API ref: https://github.com/knocklabs/knock-node/blob/f08210b60707f862a21de6b15d1856c3ce47e262/src/resources/messages/interfaces.ts#L64

### DIFF
--- a/content/reference.mdx
+++ b/content/reference.mdx
@@ -825,7 +825,7 @@ A message is a notification delivered on a particular channel to a user.
   <Attribute
     name="status"
     type="string"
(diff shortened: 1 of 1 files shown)


cellomatt (approved):
Good catch. I also cross-referenced the enum for delivery statuses in the API code and it should be as you've updated it here 👏

PR #761, opened 2025-01-28
chore(kno-7451): update docs to use workflow trigger wording instead of event action mappings

meryldakin:
### Description
Uses the "workflow trigger" wording instead of event action mappings and also removes instructions on how to commit them.

### Tasks
[kno-7451](https://linear.app/knock/issue/KNO-7451/docs)


cellomatt (approved):
🚀

PR #778, opened 2025-02-24
chore: remove APNS cert update callout

cellomatt:
### Description

This PR removes the callout on our APNS docs regarding the Apple cert update requirement, now that the deadline for that update has passed.


JEverhart383 (approved):
LGTM

PR #799, opened 2025-03-18
chore: add sms template overrides to mapi

cellomatt:
### Description

Add [SMS settings and overrides](https://docs.knock.app/integrations/sms/settings-and-overrides) to the mAPI reference.

### Screenshots

image


cellomatt:
This will be replaced shortly by generated reference

PR #800, opened 2025-03-20
feat: 1-click unsubscribe docs

connorlindsey:
### Description
Adds 1-click unsubscribe docs

### Todos
- [x] Add to broadcast docs

### Tasks
[KNO-7966](https://linear.app/knock/issue/KNO-7966/[docs]-1-click-unsubscribe-link-docs)

### DIFF
--- a/content/concepts/broadcasts.mdx
+++ b/content/concepts/broadcasts.mdx
@@ -91,7 +91,7 @@ Additionally, developer tools and extensions are available for broadcasts as the
 
 Broadcasts respect all [user preferences](/concepts/preferences) that are configured in Knock. That means for any categories added to a broadcast, preferences linked to those categories will be respected. Additionally, any channel type preferences set for a user will be respected during the execution of a broadcast.
 
-We're also working on a new set of promotional preference settings that will automatically add a 1-click unsubscribe header and link to your email messages, to respect CAN-SPAM compliance. This will be available in the coming weeks.
+Knock also supports [commercial unsubscribe](/preferences/commercial-unsubscribe) preferences that will automatically add a 1-click unsubscribe header and link to your email messages, to respect CAN-SPAM compliance.
 
 ## Frequently asked questions
 


content/preferences/overview.mdx
```diff
@@ -15,6 +15,8 @@ When Knock runs a workflow for a user, we evaluate their `PreferenceSet`. A mess
 
 A preference set is built using three keys: `categories`, `channel_types`, `workflows`. These keys resolve to boolean values to determine if a user has opted out of receiving a notification.
 
+The default preference set also has the `commercial_unsubscribe` which determines if the recipient should receive notifications sent by commercial workflows or broadcasts.
```
cellomatt, line 18:
```suggestion
The default preference set also has a `commercial_unsubscribe` key which determines if the recipient should receive notifications sent by commercial workflows or broadcasts.
```

cellomatt: I wonder if we should include an example below? Also I think there are a few things we could disambiguate here:

- The term "default", since there's the environment-level "default preference set" in addition to a recipient's `default` preferences.
- Whether we add this key in automatically (i.e. `"commercial_unsubscribe": true`) to each `PreferenceSet` that is set in Knock, regardless of whether it is provided by the customer's application? 
- Does `true` = send and `false` = don't send as usual? I think we should probably say this explicitly, because to me a setting of `"commercial_unsubscribe": true` reads like "it is true that I have unsubscribed, therefore I will not receive a message" but I think it's actually the opposite based on how our preferences model works.

content/preferences/commercial-unsubscribe.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+Knock provides built-in support for commercial email unsubscribe functionality, allowing recipients to opt out of promotional or commercial messages with a single click.
+
+## How commercial unsubscribe works
+
+When you mark a workflow or broadcast as commercial, Knock automatically handles the necessary unsubscribe functionality:
+
+1. Adds required unsubscribe headers to all emails sent through that workflow or broadcast.
+1. Provides an unsubscribe URL variable that can be included in your email templates.
+1. Manages recipient opt-outs during preference set evaluation.
```
cellomatt, line 16:
```suggestion
1. Manages recipient opt-outs during [preference set evaluation](/preferences/overview#preference-evaluation-rules).
```
Maybe a backlink to https://docs.knock.app/preferences/overview#preference-evaluation-rules here? And might be helpful to include some additional information about commercial unsubscribe in that section as well.

content/preferences/commercial-unsubscribe.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+## Configuring commercial workflows
+
+To enable commercial unsubscribe functionality for a workflow or broadcast:
+
+1. Navigate to the workflow or broadcast.
+1. For a workflow, click "Manage workflow." For a broadcast, click "Edit details."
+1. Toggle "Commercial."
+1. Save your changes.
```
cellomatt, line 25: Does this setting exist outside of commit model, or should we call that out too?

content/preferences/overview.mdx
cellomatt: I see we include some of this on the other page under "Preference evaluation rules". I think it doesn't hurt to include at least some of the info in both places, in case someone reads the preferences overview page without looking further into commercial unsub. Perhaps linking from here to `/preferences/commercial-unsubscribe#preference-evaluation-rules` is enough

content/preferences/commercial-unsubscribe.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+When a user unsubscribes by clicking the unsubscribe link, Knock displays a confirmation page showing they have been successfully unsubscribed from commercial messages. You can customize this page by navigating to **Developers** > **Preferences**, then clicking the **Unsubscribe** tab.
+
+### Standard confirmation page
+
+You can customize the title and body text that will appear on the Knock confirmation page.
+
+### Custom redirect URL
+
+You can provide a URL that recipients should be redirected to after unsubscribing.
```
cellomatt, line 61: IMO, we could maybe cut these two subheaders completely and mention what is customizable under the "Configuring the confirmation page header" (i.e. without having separate sections), or we should include screenshots under these subheaders.

content/preferences/commercial-unsubscribe.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+## Preference evaluation rules
+
+When a recipient clicks the unsubscribe link, their default preference set will be updated, marking `commercial_unsubscribe` as true. They will be opted-out of commercial messages, and they will continue to receive transactional messages based on their other preferences.
```
cellomatt, line 65:
```suggestion
When a recipient clicks the unsubscribe link, their `default` preference set will be updated, marking `commercial_unsubscribe` as `true`. They will be opted-out of commercial messages, and they will continue to receive transactional messages based on their other preferences.
```

cellomatt: Question: how do we handle this for tenant-specific preferences? What happens if a `tenant` is applied to a given workflow trigger (either when the email that has the unsub link in it is generated, or after they have clicked it on a different workflow run's email)?

cellomatt: Hmmm. I will run a test here, but I _think_ (preferences get really confusing) that if a given recipient has a `PreferenceSet` for a given `tenant.id`, we don't fall back to that user's `default`: https://docs.knock.app/preferences/tenant-preferences#tenant-preference-evaluation-rules

In other words, if a `tenant` is applied to a workflow trigger, we evaluate any default preferences that are set on that tenant, PLUS we then merge in the user's own tenant-specific preferences. But the tenant-level default overrides the recipient's `default`

cellomatt: Ok resolving this because we've added a callout on how this will work and resolving product behavior separately 👍

cellomatt (approved):
This lgtm! I incorporated feedback from first round and we've also addressed outstanding questions on functionality. Might be good to get @rachael-t to take one final look before we ship but I think this is g2g 🚀

content/preferences/commercial-unsubscribe.mdx
```diff
@@ -0,0 +1,98 @@
+---
+title: "Commercial Unsubscribe"
```
rachael-t, line 2:
```suggestion
title: "Commercial unsubscribe"
```
Sentence casing

rachael-t (approved):
🤩

PR #830, opened 2025-04-11
feat(KNO-8385): add example JSON for MS Teams Adaptive Cards

mattmikolay:
### Description

It’s not clear from our docs what the JSON payload should look like when a user wants to send a custom Adaptive Card to Microsoft Teams, so this PR adds an example.

### Tasks

[KNO-8385](https://linear.app/knock/issue/KNO-8385/update-ms-teams-docs-to-better-describe-using-custom-adaptive-cards)


cellomatt (approved):
thank you!

PR #849, opened 2025-04-25
More performant sidebar

MikeCarbone:
The open / close and scroll functionality led to some jank, especially on slower machines. This PR reduces the amount of calls we make on scroll and leads to a much smoother sidebar UX

### DIFF
--- a/components/ui/Page/Sidebar.tsx
+++ b/components/ui/Page/Sidebar.tsx
@@ -112,6 +112,7 @@ const Item = ({
           },
         }
       : {
+          px: "1",
           color: "gray",
         };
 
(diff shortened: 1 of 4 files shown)


components/ui/Page/Sidebar.tsx
```diff
@@ -112,6 +123,7 @@ const Item = ({
           },
         }
       : {
+          px: "1",
```
MikeCarbone, line 126: menu was nesting incorrectly

components/ui/Page/Sidebar.tsx
```diff
@@ -46,6 +47,17 @@ const Item = ({
...
 
+  // Create the debounced function once when component mounts
+  // This helps produce a smoother experience when scrolling fast
+  const debouncedHighlight = useMemo(
+    () =>
+      debounce((path: string) => {
+        setIsOpen(true);
+        highlightResource(path);
+      }, 300), // The lower the number here, the quicker the highlight, but can get laggy if too low
+    [], // Empty dependency array means this is only created once
```
cjbell, line 58: Shouldn't this be bound to `path`? So it reruns every time path changes?

MikeCarbone: Nope, path is passed in via the function parameter!

PR #865, opened 2025-04-29
feat(kno-8533) replace all sdk examples with new sdk format

meryldakin:
### Description

Updates all of the code snippets to the new SDK import and function signatures.

I relied HEAVILY on cursor for the code snippets that I'm unfamiliar with (Go & Java). If there's anyone who has familiarity there and can spot check, that would be very helpful.

### Tasks
[KNO-8533: \[docs\] replace all sdk examples with new sdk format](https://linear.app/knock/issue/KNO-8533/[docs]-replace-all-sdk-examples-with-new-sdk-format)


cjbell (approved):
we can merge this as is, but i think it would be much better for us to remove any of our `data/code/**` snippets and replace them with the examples in the documented/generated openapi spec. I know it's not a perfect 1:1 but that feels like a much more maintainable long term option.

PR #870, opened 2025-04-30
fix(kno-8549): fix message status schemas

meryldakin:
### Description
Message status responses weren't showing because they reference "allOf" the Message schema in the OpenApi spec. 

### Tasks
[KNO-8549: \[docs\] fix mark messages status schemas](https://linear.app/knock/issue/KNO-8549/[docs]-fix-mark-messages-status-schemas)

### DIFF
--- a/data/specs/api/stainless.yml
+++ b/data/specs/api/stainless.yml
@@ -45,69 +45,69 @@ resources:
       inline_identify_user_request: '#/components/schemas/InlineIdentifyUserRequest'
     methods:
       get: get /v1/users/{user_id}
-      list: 
+      list:
         endpoint: get /v1/users
       update: put /v1/users/{user_id}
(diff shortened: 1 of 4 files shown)


data/specs/api/stainless.yml
```diff
@@ -303,29 +303,29 @@ resources:
...
     reverse_domain: app.knock.api
-    production_repo: null
+    production_repo: knocklabs/knock-java
     publish:
       maven: false
     skip: false
   ruby:
     gem_name: knockapi
-    production_repo: null
+    production_repo: knocklabs/knock-ruby
```
MikeCarbone, line 328: including these here was intentional?

meryldakin: it is, this is just the latest from the generated specs!

PR #882, opened 2025-05-01
fix(kno-8569): stop scrolling when typing in search bar

meryldakin:
### Description

**Problem:**
When the page is scrolled down from the top, typing into the global search input located in the PageHeader causes the page viewport to incrementally scroll upwards with each keystroke. This behavior was observed across multiple browsers (Chrome, Safari).

**Root Cause:**
The root cause was identified as the global CSS rule `scroll-padding-top: var(--top-padding);` applied to the html element in styles/index.css. This property, intended to add padding above targets of fragment identifier navigation (# links) to prevent them from being obscured by sticky headers, was being incorrectly applied by browsers during input focus events. When typing into the search input while the page was scrolled, the browser attempted to ensure the focused input respected this top padding, resulting in the incremental upward scroll on each keystroke.

**Solution:**
The scroll-padding-top: var(--top-padding); rule was moved from the html selector to the body selector in styles/index.css.
Result:

The intended behavior of scroll-padding-top for fragment identifier navigation (clicking # links) is preserved, as the body element typically acts as the main scroll container for page content containing such anchor targets. This ensures headings or sections linked via # are still correctly positioned below the sticky header.

### Tasks

[KNO-8569: \[docs\] fix scrolling when typing in search box](https://linear.app/knock/issue/KNO-8569/[docs]-fix-scrolling-when-typing-in-search-box)

### DIFF
--- a/components/ui/Page/helpers.ts
+++ b/components/ui/Page/helpers.ts
@@ -41,9 +41,15 @@ export const highlightResource = (
 
     // Scroll to the content on the page
     if (newActiveItem) {
-      newActiveItem.scrollIntoView({
-        behavior: "instant",
-        block: "start",
+      // Get the desired padding from the CSS variable
+      const topPadding = parseInt(getComputedStyle(document.documentElement).getPropertyValue('--top-padding'), 10) || 0;
+      const elementRect = newActiveItem.getBoundingClientRect();
+      const targetScrollY = window.scrollY + elementRect.top - topPadding;
+
+      window.scrollTo({
+        top: targetScrollY,
+        // Use 'instant' matching the original scrollIntoView behavior here
+        behavior: "instant", 
       });
 
       newActiveItem.focus();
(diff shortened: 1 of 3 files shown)


MikeCarbone:
Breaks jump links on API reference, so we might be best just removing this scroll padding thing altogether and accounting for it when manually navigating: [https://www.loom.com/share/347455ae2dac4183b50438f25abff768](https://www.loom.com/share/347455ae2dac4183b50438f25abff768)

Also when we navigate directly to a URL its affected: [https://docs-git-md-kno-8569-knocklabs.vercel.app/api-reference/tenants/set](https://docs-git-md-kno-8569-knocklabs.vercel.app/api-reference/tenants/set)

So would have to adjust the `useInitialScrollState` hook in `components/ui/Page/helpers.ts`

components/ui/Page/helpers.ts
```diff
@@ -118,7 +129,26 @@ export const useInitialScrollState = () => {
...
+            ),
+            10,
+          ) || 0;
+        const elementRect = element.getBoundingClientRect();
+        // Calculate scroll position relative to the current scroll, adjusted for padding.
+        // No need for window.scrollY here as we are setting the absolute scroll position on load (within timeout).
+        const targetScrollY = window.pageYOffset + elementRect.top - topPadding;
+
+        // Use 'auto' for immediate jump on load, respecting user agent settings for motion if any.
+        window.scrollTo({ top: targetScrollY, behavior: "auto" });
```
MikeCarbone, line 147: can we move this into its own function?

MikeCarbone (approved):
Nice! I wasn't able to test search on preview link but clicking around & refreshing feels great. Left one small comment

PR #890, opened 2025-05-02
fix: redirecting directly to SDK

MikeCarbone:
Fixing an issue where going directly to an SDK URL would 404

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -282,6 +282,11 @@ const nextConfig = {
       },
       // Reference redirects
       {
+        source: "/sdks/:sdk",
+        destination: "/in-app-ui/:sdk/sdk/overview",
+        permanent: true,
+      },
+      {
         source: "/sdks/:sdk/reference",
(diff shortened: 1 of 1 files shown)


meryldakin (approved):

next.config.js
```diff
@@ -281,6 +281,11 @@ const nextConfig = {
         permanent: true,
       },
       // Reference redirects
+      {
+        source: "/sdks/:sdk",
+        destination: "/in-app-ui/:sdk/sdk/overview",
```
cellomatt, line 286:
```suggestion
        destination: "/in-app-ui/:sdk/overview",
```
I think it should be like this? i.e. https://docs.knock.app/in-app-ui/react/overview

MikeCarbone: first `:sdk` is the SDK name, like "react" or "android"

second sdk is just "sdk"
```
https://docs.knock.app/in-app-ui/:sdk/sdk/overview
=== matches ===
https://docs.knock.app/in-app-ui/react/sdk/overview
```

PR #896, opened 2025-05-02
fix: build speeds

MikeCarbone:
- Cuts build times from 12 minutes to 6 minutes (**100% improvement!** 🎉 )
- Improves search indexing compatibility
- Shuts up hydration warnings caused by `useLayoutEffect`

### DIFF
--- a/components/ui/ApiReference/ApiReference.tsx
+++ b/components/ui/ApiReference/ApiReference.tsx
@@ -38,6 +38,7 @@ function ApiReference({
       <Meta
         title={`${name} | Knock Docs`}
         description={`Complete reference documentation for the Knock ${name}.`}
+        canonical={`/${basePath}`}
       />
       <TelegraphPage.Masthead title={name} />
       <ApiReferenceProvider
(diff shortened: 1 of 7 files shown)


components/ui/ApiReference/ApiReference.tsx
```diff
@@ -38,6 +38,7 @@ function ApiReference({
       <Meta
         title={`${name} | Knock Docs`}
         description={`Complete reference documentation for the Knock ${name}.`}
+        canonical={`/${basePath}`}
```
MikeCarbone, line 41: We need to make sure canonical points to the root of the page.
So `/api-reference/messages/get` has a canonical of `/api-reference`

Without this, our site will get penalized for duplicate content, as google will see "all of these unique URLs & pages have the same thing!", canonical protects against that

WorkOS does this too. Check out the canonical link in the `<head>`:
https://workos.com/docs/reference/user-management/session-helpers/load-sealed-session

components/ui/Page/Sidebar.tsx
```diff
@@ -112,7 +111,7 @@ const Item = ({
     [isSearchOpen], // Empty dependency array means this is only created once
   );
 
-  useLayoutEffect(() => {
+  useEffect(() => {
```
MikeCarbone, line 114: Still works but nextjs was yelling at us so much for this causing hydration issues

cjbell (approved):

PR #903, opened 2025-05-05
chore: update example code

JEverhart383:
### Description

Updates slack ex. code to show how to pass additional scopes


content/integrations/chat/slack/sending-a-direct-message.mdx
```diff
@@ -124,7 +124,10 @@ Here's an example of how to use them:
 <SlackAuthButton
     slackClientId="Slack app client ID"
     redirectUrl="The URL of your application to return to once Slack authorization is complete"
-    additionalScopes={["users:read", "users:read.email"]}
+    options={{
```
cellomatt, line 127: If I'm reading [this](https://github.com/knocklabs/javascript/pull/522/files) correctly, it looks like for the `SlackAuthButton` we are just adding an additional `scopes` prop and `options` only changes on the `useSlackAuth` hook?

PR #911, opened 2025-05-12
feat(KNO-8528): improve select and multi-select field sections of message types doc

mattmikolay:
### Description

It’s not immediately clear how to access selected values in a template preview when using select and multi-select fields, so this PR updates the message types doc to improve the descriptions and examples.

I also updated the **Variants** section to mention that a variant schema with the key `"default"` is required.

### Tasks

[KNO-8528](https://linear.app/knock/issue/KNO-8528/docs-update-message-type-schema-docs-to-better-describe-how-to-use)

### DIFF
--- a/content/in-app-ui/message-types.mdx
+++ b/content/in-app-ui/message-types.mdx
@@ -403,8 +403,8 @@ A plain text, single line text field.
 
 **Settings**
 
-- `required` (boolean): indicates this field is required
-- `description` (string): an optional friendly description
+- `required` (boolean): Indicates this field is required
+- `description` (string): An optional friendly description
 - `default` (string): The default value to display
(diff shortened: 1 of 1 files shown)


MikeCarbone (approved):

mattmikolay:
> Looks good. One nit before we ship. We're using inconsistent casing after the list of settings. Not introduced with this pr but would be good to clean up while we're in here.

Good catch! I just made the casing consistent. I’m going to merge, but I’m also wondering if we would prefer to show the field settings as a table, rather than a bulleted list, similar to how we display [channel data requirements](https://docs.knock.app/managing-recipients/setting-channel-data#provider-data-requirements):

Screenshot 2025-05-13 at 9 49 17 AM

Thoughts, @samseely?

PR #929, opened 2025-05-19
fix: update node sdk migration guide info on identifying users

connorlindsey:
### Description

Update node sdk migration guide info on identifying users

### DIFF
--- a/content/developer-tools/migration-guides/node.mdx
+++ b/content/developer-tools/migration-guides/node.mdx
@@ -160,31 +160,9 @@ await client.users.identify("jhammond", {
 **1.x release:**
 
 ```javascript
-await client.users.identify("jhammond", {
+await client.users.update("jhammond", {
   name: "John Hammond",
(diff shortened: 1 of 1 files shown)


content/developer-tools/migration-guides/node.mdx
```diff
@@ -160,7 +160,7 @@ await client.users.identify("jhammond", {
 **1.x release:**
 
 ```javascript
-await client.users.identify("jhammond", {
+await client.users.update("jhammond", {
   name: "John Hammond",
   email: "jhammond@ingen.net",
 });
```
connorlindsey, line 166: Should we remove the objects section below since they're the same between versions?

JEverhart383: I think we could since this feels kind of redundant. I know some of the params have changed between versions, so if there's some 'more extended' operation, we could compare the two, but that may not apply here. I'm thinking of like changes to how we pass the idempotency key, for example

PR #942, opened 2025-05-23
Sts feed doc improvements

samseely:


data/sidebars/inAppSidebar.ts
```diff
@@ -25,10 +25,26 @@ export const IN_APP_UI_SIDEBAR: SidebarSection[] = [
         slug: "/feeds",
         pages: [
           { slug: "/overview", title: "Overview" },
+          {
+            slug: "/styling",
+            title: "Styling",
+          },
+          {
+            slug: "/custom-ui",
+            title: "Custom UI (headless)",
```
rachael-t, line 34: Wondering if we should name this to just "Headless"?

data/sidebars/inAppSidebar.ts
```diff
@@ -112,20 +128,12 @@ export const REACT_SIDEBAR: SidebarSection[] = [
...
-      { slug: "/feed", title: "Notification feed" },
-      { slug: "/toasts", title: "Toasts" },
-      { slug: "/inbox", title: "Notification inbox" },
-      {
-        slug: "/custom-notifications-ui",
-        title: "Custom feed UI (headless)",
-      },
+      { slug: "/feed", title: "Feed" },
+      { slug: "/toasts", title: "Toast" },
+      { slug: "/inbox", title: "Inbox" },
```
rachael-t, line 133: I really like having these titles cleaned up.

data/sidebars/inAppSidebar.ts
samseely: I thought about that but did some research and felt like not everyone knows what headless means so better to be explicit and say what it is

data/sidebars/inAppSidebar.ts
samseely: I agree! soon we'll add modal, banner, etc

cjbell (commented):
One comment on the custom UI page

PR #950, opened 2025-05-28
feat: allow sidebar to not be nested

cjbell:
### Description

This PR allows us to have non-nested sidebar items, so that we don't need to wrap "Developer tools" under a developer tools section in the sidebar. It does this by introducing a new recursive `SidebarContent` type, which can then have subpages (or subpages of subpages).

Most of the changes here are reworking the breadcrumbs as those were most borked because of this. It could definitely do with another pass later, but I think this should suffice for now. 

I implemented the changes in the following places:

- Integrations > Overview
- Developer tools
- Tutorials

### Screenshots

image

### DIFF
--- a/components/ui/Page/helpers.ts
+++ b/components/ui/Page/helpers.ts
@@ -129,10 +129,6 @@ export const updateNavStyles = (resourceUrl: string) => {
   activeElements.forEach((element) => {
     if (element.dataset.resourcePath !== resourceUrl) {
       element.setAttribute("data-active", "false");
-      // For some reason, the color persists and we have to set it back to default by hand
-      if (element.firstChild) {
-        (element.firstChild as HTMLElement).style.color = "var(--tgph-gray-11)";
-      }
     }
   });
 };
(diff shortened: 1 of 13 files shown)


components/ui/Page/helpers.ts
```diff
@@ -129,10 +129,6 @@ export const updateNavStyles = (resourceUrl: string) => {
   activeElements.forEach((element) => {
     if (element.dataset.resourcePath !== resourceUrl) {
       element.setAttribute("data-active", "false");
-      // For some reason, the color persists and we have to set it back to default by hand
-      if (element.firstChild) {
-        (element.firstChild as HTMLElement).style.color = "var(--tgph-gray-11)";
-      }
```
MikeCarbone, line 135: Had this here to address the colors sticking on API reference sidebars when you click around

Screenshot 2025-05-28 at 12 27 53 AM

data/sidebars/developerToolsSidebar.ts
```diff
@@ -1,79 +1,81 @@
...
 
-export const DEVELOPER_TOOLS_SIDEBAR: SidebarSection[] = [
+const baseSlug = "/developer-tools";
+
+export const parentSection = {
+  slug: baseSlug,
+  title: "Developer tools",
+};
+
+export const DEVELOPER_TOOLS_SIDEBAR_CONTENT: SidebarContent[] = [
```
MikeCarbone, line 10:
```suggestion
export const DEVELOPER_TOOLS_SIDEBAR: SidebarContent[] = [
```
nit but the other sidebars don't use the `_CONTENT` suffix. You can keep if you want nbd

PR #958, opened 2025-06-02
fix(kno-8744): more consistent API linking

MikeCarbone:
[[Docs] Bug in redirect from search to API Reference](https://linear.app/knock/issue/KNO-8744/docs-bug-in-redirect-from-search-to-api-reference)

This PR makes navigating to API reference pages more consistent.

The problem was that we'd try to navigate to the section prematurely, as the page was still rendering (API references are big pages, so this takes time). I added a check to make sure the page was at its final height before navigating.

Also, the sidebar wouldn't open in Scoti's video because SlackChannelData is a deeply nested page. Sidebar open state code didn't account for pages that were nested this deeply, so I made the function recursive to handle infinitely nested pages.

### DIFF
--- a/components/ui/Page/helpers.ts
+++ b/components/ui/Page/helpers.ts
@@ -1,5 +1,5 @@
 import { useRouter } from "next/router";
-import { useLayoutEffect } from "react";
+import { useLayoutEffect, useState } from "react";
 import { debounce } from "@/lib/debounce";
 
 export const stripPrefix = (path: string) => {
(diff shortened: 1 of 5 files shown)


components/ui/Page/helpers.ts
```diff
@@ -144,28 +144,59 @@ const scrollToElementWithPadding = (document: Document, element: Element) => {
...
-        `[data-resource-path="${resourcePath}"]`,
-      );
-
-      if (element) {
-        scrollToElementWithPadding(document, element);
+    let prevHeight = document.body.scrollHeight;
+    let stableCount = 0;
+    const maxStableChecks = 3; // Require 3 consecutive stable measurements
+    let interval: NodeJS.Timeout;
+    interval = setInterval(() => {
```
mattmikolay, line 163: Nit: use a `const` for `interval` since you never re-assign it?

mattmikolay (approved):

PR #986, opened 2025-06-16
chore: upgrade next to 14

cjbell:
### Description

Updates next to 14 per issue in Vanta


MikeCarbone:
Just waiting for the build to check everything out. We can close #699 after this too

PR #991, opened 2025-06-16
chore(KNO-8906): remove @analytics/segment dependency

mattmikolay:
### Description

[`@analytics/segment`](https://github.com/DavidWells/analytics/tree/master/packages/analytics-plugin-segment) was added by #500, but I’m pretty sure this dependency is entirely unused. That same PR also installed and set up [`@segment/snippet`](https://github.com/segmentio/snippet).

This PR removes `@analytics/segment` from the dependencies, to resolve [this automated Dependabot alert](https://github.com/knocklabs/docs/security/dependabot/30).

### Tasks

[KNO-8906](https://linear.app/knock/issue/KNO-8906/docs-remove-analyticssegment-dependency)

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,6 @@
   "dependencies": {
     "@algolia/autocomplete-js": "^1.6.3",
     "@algolia/autocomplete-theme-classic": "^1.6.3",
-    "@analytics/segment": "^2.1.0",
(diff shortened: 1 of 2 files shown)


MikeCarbone (approved):
nice!

PR #1004, opened 2025-06-24
Guides cli docs

connorlindsey:
### Description
Adds info for guides CLI commands

### Screenshots
CleanShot 2025-06-24 at 12 49 36


rachael-t (approved):
Small nit that the double "Note:" and "**Note:**" is visually a little confusing, but I recognize that this is temporary while Guides is still in Beta, so I'm fine with leaving it as is for now.
Screenshot 2025-06-24 at 2 12 25 PM

PR #1014, opened 2025-07-03
chore: upgrade telegraph icon, migrate icons, make lucide usage static

kylemcd:
### Description
- Upgrades `@telegraph/icon` and other telegraph packages to the latest versions.
- Migrates usage of `Lucide` to new import strategy.
- Updates `Icons.tsx` to utilize named exports for icons to decrease bundle size.
- Updates `Blocks.tsx` to utilize typesafe `icon` variable. Also, updates function to throw an error when an invalid icon is used so the build fails in these cases rather than a runtime error occurring.

### Tasks
KNO-9052


components/ui/ApiReference/SchemaProperties/PropertyRow.tsx
```diff
@@ -134,7 +134,7 @@ const ExpandableButton = ({ children, isOpen, onClick }) => (
       marginLeft: "calc(var(--tgph-spacing-1) * -1)",
     }}
     icon={{
-      icon: Lucide.ArrowRight,
+      icon: ChevronRight,
```
cellomatt, line 137: Just making sure `ArrowRight` -> `ChevronRight` is intended

cellomatt (approved):
Amazing ty!

components/ui/ApiReference/SchemaProperties/PropertyRow.tsx
kylemcd: Ope nope, good catch ty

components/ui/Icons.tsx
```diff
@@ -100,5 +121,32 @@ export const Icons = {
...
+  Terminal,
+  Landmark,
+  Infinity,
+  Unplug,
+  Code,
+  ChevronsLeftRightEllipsis,
+  Package,
 };
+
+export const Icons = Object.entries(PascalCaseIcons).reduce(
```
MikeCarbone, line 146: One thing to call out here is I had a system for referencing Icons via string in components. Looking for example to make sure this is okay

MikeCarbone (approved):
Clicked around too and icons look good!

PR #1017, opened 2025-07-07
Sam guide component docs

samseely:
- add image to feed docs
- add initial pages for modal, banner, card


content/in-app-ui/react/modal.mdx
```diff
@@ -0,0 +1,259 @@
...
+          </KnockGuideProvider>
+        </KnockProvider>
+      );
+    };
+    ```
+
+  </Step>
+  <Step title="Add the Modal component">
+    
+    Import Knock's pre-built `Modal` component and place it under the `KnockGuideProvider` near the top of your application. The `<Modal/>` component is most effective when placed in the root layout of your application, this ensures that your modal is available to render on every page of your application.
```
thomaswhyyou, line 95:
```suggestion
    Import Knock's pre-built `Modal` component and place it inside the `KnockGuideProvider` near the top of your application's component tree. The `<Modal/>` component is most effective when placed in the root layout of your application, this ensures that your modal is available to render on every page of your application.
```

content/in-app-ui/react/modal.mdx
```diff
@@ -0,0 +1,259 @@
...
+        <KnockProvider
+          apiKey={process.env.NEXT_PUBLIC_KNOCK_API_KEY}
+          userId={currentUser.id}
+        >
+          <KnockGuideProvider
+            channelId={process.env.NEXT_PUBLIC_KNOCK_GUIDE_CHANNEL_ID}
+          >
+            <Header />
+            <Sidebar />
+            {/* Rest of your app */}
```
thomaswhyyou, line 85:
```suggestion
            <MyLayout>
              {/* Rest of your app */}
            </MyLayout>
```
Just a thought that I had, whether it'd be good to reference the component we are using as an example below.

content/in-app-ui/react/modal.mdx
```diff
@@ -0,0 +1,259 @@
...
+
+  </Step>
+  <Step title="Add the Modal component">
+    
+    Import Knock's pre-built `Modal` component and place it under the `KnockGuideProvider` near the top of your application. The `<Modal/>` component is most effective when placed in the root layout of your application, this ensures that your modal is available to render on every page of your application.
+
+    ```tsx title="Add the modal near the top of your application."
+    import { Modal } from "@knocklabs/react";
+
+    const DashboardLayout = ({children}) => {
```
thomaswhyyou, line 100:
```suggestion
    const MyLayout = ({children}) => {
```

content/in-app-ui/react/modal.mdx
```diff
@@ -0,0 +1,259 @@
...
+- **Archived.** The modal has been archived (dismissed) by the user.
+
+## Styling your modal
+
+### CSS customization
+
+You can style your modal using standard CSS. Here's a basic example:
+
+```css title="Modal styles"
+.modal-overlay {
```
thomaswhyyou, line 148: The class names here are not quite right. [Here](https://github.com/knocklabs/javascript/blob/18be7a567363c1056b6f420aa45c6d83643f49ed/packages/react/src/modules/guide/components/Modal/styles.css) is what we have in the latest SDK we can reference. 

One other thing to touch on in this "Styling" section is that we use [css variables](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascading_variables/Using_CSS_custom_properties) (e.g. `--knock-guide-accent`) in our css definitions (linked above). This was intentional as we wanted to make it easy to "theme" the prebuilt components using your own color tokens if you are otherwise happy with how the components are styled. More specifically, the ones prefixed with `--knock-guide-` are meant to be overridden for the theming purposes. 

So there are different layers of abstractions we provide that you can use to approach styling, from the simplest (but limited) to most involved (but flexible):
* Override the `--knock-guide-` css vars to "theme" the components with your own colors
* Override some or all of the css definitions to completely customize the styles
* Use the individual [sub components](https://github.com/knocklabs/javascript/blob/18be7a567363c1056b6f420aa45c6d83643f49ed/packages/react/src/modules/guide/components/Modal/Modal.tsx#L339-L351) to compose your own component (and use inline style props, if needed)
  * This is technically available but i think at this point most people would reach for the BYOC option
* "Bring your own component (BYOC)" and style freely, and leverage the useGuide hook to power it.

content/in-app-ui/react/modal.mdx
```diff
@@ -0,0 +1,259 @@
...
+  const { step } = useGuide({ type: "modal" });
+
+  useEffect(() => {
+    if (step) step.markAsSeen();
+  }, [step]);
+
+  if (!step) return null;
+
+  return (
+    <div onClick={() => step.markAsInteracted()}>
```
thomaswhyyou, line 234: I think it'd be better to show an example with a button and `step.markAsInteracted()` as an onClick handler for the button rather than at the root container, because this markup structure can lead to firing markAsInteracted when a dismiss button is clicked inside, for example. You can prevent that with `event.preventDefault()` and/or `event.stopPropagation()` but that's more things to explain.

```
{step.content.primary_button && (
  <button onClick={() => step.markAsInteracted()}>...</button>
)}
```

PR #1019, opened 2025-07-07
feat: enhance tracking documentation with custom domain section

samseely:
adding section on enabling custom domains to link tracking docs


rachael-t (commented):
Currently, this is an Enterprise-only offering. Should we add a callout to the top?

samseely:
> Currently, this is an Enterprise-only offering. Should we add a callout to the top?

Yep great call, updating

PR #1021, opened 2025-07-08
chore(kno-9044): add replace strategy and clarify merge hierarchy

cellomatt:
### Description

This PR updates our preferences overview page:
- Adds a section on merge hierarchy to explicitly outline the order in which various preferences take precedence when merged
- Adds a bullet point under "Advanced concepts" on setting the merge strategy
- Adds an FAQ outlining implementation and considerations for using the `replace` merge strategy

### DIFF
--- a/content/preferences/overview.mdx
+++ b/content/preferences/overview.mdx
@@ -205,7 +205,7 @@ There are four steps to building a preference center with Knock:
 
 ## Merging preferences
 
-When a default `PreferenceSet` exists for an environment or tenant, Knock will merge all applicable preferences for a recipient when evaluating whether or not to send a notification. Any preferences set at the recipient-level will take precedence in the merge.
+When a default `PreferenceSet` exists for an environment or tenant, Knock will merge all applicable preferences for a recipient when evaluating whether or not to send a notification. Any preferences set at the recipient level will take precedence in the merge, according to the hierarchy outlined below.
 
 ## Preference evaluation rules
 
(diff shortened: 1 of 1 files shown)


content/preferences/overview.mdx
```diff
@@ -216,6 +216,22 @@ When a workflow is triggered, Knock will evaluate the preferences for each `reci
     - If you do not set a preference for a given channel, workflow, or workflow category, Knock defaults them to `true`.
     - When a recipient clicks an unsubscribe link, their `default` preference set will be updated, marking `commercial_unsubscribe` as `true`. They will be opted-out of commercial messages, and they will continue to receive transactional messages based on their other preferences. Read more about commercial unsubscribe [here](/preferences/commercial-unsubscribe).
   </Accordion>
+  <Accordion title="Merge hierarchy">
```
rachael-t, line 219: I've been going back and forth on this, but initially, it felt weird to read in the section above "according to the hierarchy outlined below" and then have this accordion in a different section. I'm not sure if it would make more sense to move this specific accordion up to the "Merging preferences" section? I understand why you put it under "Preference evaluation rules" but it felt a little disorganized to me initially.

rachael-t:
I have some questions about the layout and would be interested to hear what others think. Overall though I'm so thankful that you took the time to document these missing pieces that have been coming up more often 🙏

content/preferences/overview.mdx
samseely: Yep had same exact experience reading through it for first time. When I reached "hierarchy outlined below", I expected to see it next, but when I saw another H2 i thought something was missing.

content/preferences/overview.mdx
```diff
@@ -216,6 +216,22 @@ When a workflow is triggered, Knock will evaluate the preferences for each `reci
...
+    The following hierarchies are used when merging preferences (in order of highest precedence to lowest):
+
+    **When no `tenant` is applied to the workflow trigger:**
```
samseely, line 222: This part is good but I wonder if there's a diagram we could put together to try and visualize this a bit more. 

Thinking something like 

CleanShot 2025-07-09 at 22 23 48@2x

samseely: Or this to try and put different levels of merge hierarchy on same visual plane
CleanShot 2025-07-09 at 22 24 33@2x

content/preferences/overview.mdx
```diff
@@ -326,4 +343,72 @@ You can update the preferences of up to 1000 users in a single batch by using th
     ```
 
   </Accordion>
+  <Accordion title="What is the merge strategy, and how can I use it to override the merge hierarchy?">
```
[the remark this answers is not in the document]
samseely: I actually think okay to place it here. It is a lot of information, but it's enough of an advanced use case that i think to put it in main TOC of page would feel like too much. 

That said I do think we should include a visual aid (similar to what I put on comment above) that shows how the merge hierarchy changes with merge strategy. 

Just think with multi-layer hierarchies, doesn't hurt to have it in writing and in visuals to help people grok it

samseely (commented):
@cellomatt I really appreciate you taking learnings from Multiverse and bringing into docs in a way that enables future customers to have a better experience. That is what it's all about. ❤️

Had a few thoughts reading doc, mainly around trying to include some diagrams. Nothing else from me so feel free to merge once we get some of that in there!

content/preferences/overview.mdx
cellomatt: Ok thanks! This is good feedback, I actually had it in the merge section at first but moved it because it's evaluated at runtime. But I think it does make more sense under the merge title. 🙏

scoti-knock (approved):

rachael-t (approved):
I really like these additions! Thanks for taking the time to address the feedback.

PR #1031, opened 2025-07-14
launchdarkly + knock tutorial doc

samseely:
- focus here is on using launchdarkly w what's available in Knock workflow engine today
- does not address using LD with guides (think that would probably be a segment to audience integration?)
- does not speak to LD<>Knock integration in future where flag eval auto-stored on knock users/objects


content/tutorials/launchdarkly-knock.mdx
```diff
@@ -0,0 +1,289 @@
...
+    user_name: user.name,
+    variant: messagingVariant,
+  },
+});
+```
+
+</Step>
+</Steps>
+
+### Method 2: Branch-based message variants
```
cjbell, line 127: I think we need to show a section where we use per-recipient trigger data to evaluate a variation per-user and then send that up in a single request to Knock, just to show how easy it is to do

content/tutorials/launchdarkly-knock.mdx
```diff
@@ -0,0 +1,289 @@
...
+- Workflow key: `welcome-sequence`
+- Add branch step with conditions:
+  - **Friendly branch**: `data.email_style == "friendly"`
+  - **Professional branch**: `data.email_style == "professional"`
+  - **Casual branch**: `data.email_style == "casual"`
+
+Configure email templates for each variant.
+
+</Step>
+<Step title="Tracking and analytics">
```
cjbell, line 263: Should we mention anything about tracking Knock messages and not just the LD tracking?

cjbell (commented):
couple of comments

cjbell (approved):

.claude/claude.md
```diff
@@ -0,0 +1,233 @@
+# Knock writing style guide
```
vlymar, line 1: does claude read this file? I thought you'd want to put it in the root dir. [These docs](https://docs.anthropic.com/en/docs/claude-code/memory#how-claude-looks-up-memories) suggest that claude will only load this file if it loads some other file from this dir

content/tutorials/launchdarkly-knock.mdx
```diff
@@ -0,0 +1,289 @@
...
+const workflowKey =
+  messagingVariant === "treatment"
+    ? "messaging-AB-test-treatment"
+    : "messaging-AB-test-control";
+
+await knock.workflows.trigger(workflowKey, {
+  recipients: [user.id],
+  data: {
+    user_name: user.name,
+    variant: messagingVariant,
```
vlymar, line 119: why include a `variant` if we're picking an entirely different workflow based on the flag variation?

.claude/claude.md
samseely: Yeah I'm not sure on that tbh. I saw we already had .claude dir and figured better to put it there since it follows same pattern as .cursor dir. 

I actually asked claude code itself about this and it said it was fine to put in .claude dir, but it might have just been agreeing with me. 

Fwiw it worked, but I'm not sure if it found claude.md because of prior context of our conversation. 

This excerpt is confusing to me: "Claude will also discover CLAUDE.md nested in subtrees under your current working directory. Instead of loading them at launch, they are only included when Claude reads files in those subtrees." It's not clear to me if this means it will automatically find claude.md no matter where it is in subtrees, or if you have to be working with that directory for it to get pulled into context.

content/tutorials/launchdarkly-knock.mdx
samseely: Yep this (and user_name) don't make sense here. Claude Code slop I missed. Thank you for catching

PR #1042, opened 2025-07-21
chore: add tags for audiences and broadcasts

cellomatt:
### Description

Our audiences and broadcast pages currently lack tags, this PR should help people find what they're looking for in docs search when they want to learn more about targeting messages to specific user cohorts and one-time sends.


samseely (approved):
Great catch!

PR #1055, opened 2025-07-25
Revision of identify docs, including addition of client-side identifi…

scoti-knock:
### Key Changes

**Added client-side identification method**
- Documents new client-side inline identification capability available in React/JavaScript SDKs
- Includes implementation examples with `KnockProvider` 
- Explains when to use client-side vs server-side identification

**Improved document structure**
- Reorganized four identification methods into logical accordion groups:
  - Explicitly identifying recipients (Individual, Bulk)
  - Inline identifying recipients (Server-side, Client-side)
- Added comprehensive FAQ section addressing common issues
- Moved preferences and channel data setting to FAQ to avoid discussion of preferences in two places and to maintain focus on identification

**Improved clarity**
- Updated endpoint list for server-side inline identification (added bulk schedules/subscriptions)
- Converted reserved properties to table format for better readability
- Fixed minor typos and tightened language where possible

**New troubleshooting guidance**
- Added FAQ for common inline identification errors (ID strings vs objects)
- Documented preferences and channel data setting during identification

### Note for review
- I meaningfully restructured the information about preferences. This includes removing the callout block about deep merging. In the new structure, I believe the concerns are immediately clear, but we should consider it would be better to maintain the callout.

### DIFF
--- a/content/managing-recipients/identifying-recipients.mdx
+++ b/content/managing-recipients/identifying-recipients.mdx
@@ -5,22 +5,21 @@ tags: ["recipients", "inline identify", "identify"]
 section: Managing recipients
 ---
 
-To send notifications to a recipient (or to reference a recipient as an actor in a notification), Knock needs the recipient's data to be stored in Knock. This process is known as **identifying** and works across user and object recipients.
+To send notifications to a recipient or reference them as an actor in a notification, their data must be stored in Knock. This process is called **identifying** and works across user and object recipients.
 
-There are three different methods to identify recipients:
+There are four different methods to identify recipients:
 
 - Individual identification
 - Bulk identification
 - Inline identification
-
-In this guide we discuss each of these methods.
+- Client-side identification
 
 <Callout
-  emoji="🌠"
+  emoji="💡"
   title="Note:"
   text={
     <>
-      all identification methods in Knock use an upsert approach. This means any
+      All identification methods in Knock use an upsert approach. This means any
       existing data for a recipient is merged during the upsert operation.
     </>
   }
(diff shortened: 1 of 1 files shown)


content/managing-recipients/identifying-recipients.mdx
```diff
@@ -5,101 +5,143 @@ tags: ["recipients", "inline identify", "identify"]
...
+              />
+            </>
+          </KnockFeedProvider>
+        </KnockProvider>
+    ```
 
-When passing a set of recipients to be inline identified, Knock will guarantee that the recipients are identified **before** any other action is executed. That makes it possible to lazily create recipients within Knock using inline identification, only creating them when needed.
+    <Callout
+      emoji="🔒"
+      title="Security considerations"
```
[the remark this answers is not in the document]
scoti-knock: good catch!

content/managing-recipients/identifying-recipients.mdx
```diff
@@ -5,101 +5,143 @@ tags: ["recipients", "inline identify", "identify"]
 section: Managing recipients
```
samseely, line 5: Unrelated but wondering if we should add a couple additional tags to this for search purposes. Two that come to mind are "users" and "objects"

content/managing-recipients/identifying-recipients.mdx
```diff
@@ -5,101 +5,143 @@ tags: ["recipients", "inline identify", "identify"]
...
 ---
 
-To send notifications to a recipient (or to reference a recipient as an actor in a notification), Knock needs the recipient's data to be stored in Knock. This process is known as **identifying** and works across user and object recipients.
+To send notifications to a recipient or reference them as an actor in a notification, their data must be stored in Knock. This process is called **identifying** and works across user and object recipients.
 
-There are three different methods to identify recipients:
+There are four different methods to identify recipients:
 
 - Individual identification
```
samseely, line 12: Here I'm wondering if we should treat this as a list with a bit more context on each one. Something like:

* **Individual identification.** Individually identify users during events when they're updated in your app (e.g. signup.)
* **Inline identification.** Identify users inline within your workflow triggers and other API calls. 
* **Bulk identification.** tktk

Other thoughts.
* Feels like bulk should go at bottom since it's the one that's used least often. 
* Is "individual identification" the right term here? Couldn't client, inline, and that one all be "individual"? If it's not the right term, what is, "one-off identification"? "Server-side identification"?

samseely: For last bullet, maybe it's "manual"? We call it "explicitly identify" below in doc. Whatever we choose think it needs to be consistent so that these bullets map to sections below.

content/managing-recipients/identifying-recipients.mdx
```diff
@@ -109,87 +151,91 @@ A recipient also accepts any number of custom properties, which are key-value pa
...
-{
-  "channel_data": {
-    "some-uuid-for-a-channel": {
-      "tokens": ["my-push-token"]
-    }
-  }
-}
-```
 
 ## When to identify recipients in Knock
```
samseely, line 156: Same thought here as on H2 above. "When to identify" is shorter, fits oneline in TOC, and "recipients in Knock" is clear from doc user is already reading

samseely (commented):
Great update, @scoti-knock! Had some other notes, less about your updates and more about existing structure/outline of page. Left comments in PR and here is a loom with my broader thoughts: https://www.loom.com/share/32cb1c90d9254b53b3d62d3ea1b7738f

content/managing-recipients/identifying-recipients.mdx
```diff
@@ -1,105 +1,153 @@
...
 
-There are three different methods to identify recipients:
+There are three ways to identify recipients:
 
-- Individual identification
-- Bulk identification
-- Inline identification
-
-In this guide we discuss each of these methods.
+- **Direct identification**. Make a single server-side API request (`users.identify`) to upsert one recipient.
```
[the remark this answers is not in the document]
scoti-knock: I think you're right on swapping these out 👍

PR #1065, opened 2025-08-04
chore: add supported destinations to warehouse sync docs

cellomatt:
### Description

This PR adds a table of Prequel-supported warehouse destination types to our docs so that our customers don't need to click through to Prequel documentation. For now, this list does not include types that are in beta status.

https://docs-git-mk-kno-9345-knocklabs.vercel.app/integrations/extensions/data-sync

Original list for comparison: https://docs.prequel.co/docs/destinations-overview#supported-destinations 

### Questions

1. Right now these are in the same order as Prequel lists them (loosely organized by "type" of destination). Would it be better to simply list them alphabetically?
2. Do we want to include destinations that are marked as "beta" on Prequel? These are theoretically available, but mileage may vary pending any unforeseen issues Prequel-side.

### DIFF
--- a/content/integrations/extensions/data-sync.mdx
+++ b/content/integrations/extensions/data-sync.mdx
@@ -33,20 +33,20 @@ We currently support the following destination database types:
 
 | Vendor            | Type           |
 | ----------------- | -------------- |
-| `snowflake`       | OLAP           |
-| `bigquery`        | OLAP           |
-| `redshift`        | OLAP           |
-| `databricks`      | OLAP           |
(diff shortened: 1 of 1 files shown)


rachael-t:
General question: are we deciding to make this change to avoid customers going to Prequel's docs to avoid any potential confusion? 

> Right now these are in the same order as Prequel lists them (loosely organized by "type" of destination). Would it be better to simply list them alphabetically?

I think it could be nice to alphabetize by each group/type

rachael-t (approved):

PR #1139, opened 2025-09-25
chore: clarify that per-recipient batching w/ timezone awareness is not currently available

cellomatt:
### Description

Updates the FAQ under the batch function to clarify that per-recipient timezone-aware batching is not currently available for early access.


rachael-t (approved):
🪟

PR #1387, opened 2026-04-10
feat(KNO-7916): Add support for linking to Accordion components

rachael-t:
### Description

Our docs rely heavily on the `Accordion` component to collapse content for better scannability. Currently, when we need to link to a specific accordion (rather than just the nearest heading), we can't - we can only link to a section. In pages with lots of accordions, like [Setting Channel Data](https://docs.knock.app/managing-recipients/setting-channel-data), this means users land on the right section but still have to hunt for the right accordion themselves.

For example, in the [MS Teams overview](https://docs.knock.app/integrations/chat/microsoft-teams/overview#how-to-set-channel-data-for-a-microsoft-teams-integration-in-knock), we link to the channel data docs for MS Teams-specific requirements but we can't deep-link directly to that accordion, so users have to find it manually.

**What this PR does**
Adds an optional `anchorId` prop to the `Accordion` component. When set, it assigns that value as the wrapper's DOM id and automatically opens the accordion on page load if the URL hash matches. Note that this is **optional** and accordions without `anchorId` should behave exactly as before (so there's no requirement to add it everywhere). Also, each `anchorId` must be unique within a page, but the same value can be reused across pages.

In this PR, I've **only** added `anchorId` to the specific example described above.

**Open questions for the team**

1. Should we do a broader audit to add `anchorId` across the docs, or just add them as we notice they'd be useful? My take is that we don't need it on every accordion but rather only where we have a stable deep link that should open a specific one. If an accordion is unlikely to be linked to directly, a heading anchor is probably enough. But open to other opinions!
2. Does the approach I've taken with `Accordion` make the most sense? I'm very open to other opinions here as well!

### Tasks

[KNO-7916](https://linear.app/knock/issue/KNO-7916/docs-support-deep-links-that-open-a-specific-accordion)

### Recordings

Current behavior
<video src="https://github.com/user-attachments/assets/ded1c809-89c3-48ca-8604-9f3acce18f44" controls width="600"></video>

New behavior
https://docs-git-rt-kno-7916-linking-to-accordions-knocklabs.vercel.app/integrations/chat/microsoft-teams/overview#how-to-set-channel-data-for-a-microsoft-teams-integration-in-knock
<video src="https://github.com/user-attachments/assets/648af771-22b2-41f5-9643-2c1807298231" controls width="600"></video>

### DIFF
--- a/components/ui/Accordion.tsx
+++ b/components/ui/Accordion.tsx
@@ -31,8 +31,8 @@ type AccordionProps = {
   title: string;
   description?: string;
   defaultOpen?: boolean;
-  /** When set, this value is used as the element `id` and the accordion opens if the URL hash matches (for deep links). */
-  anchorId?: string;
+  /** When set, this slug is used as the element `id` and the accordion opens if the URL hash matches (for deep links). Use a URL-safe hyphenated fragment, e.g. `my-section`. */
+  anchorSlug?: string;
 };
 
 // Helper function to parse title and split into text and code parts
@@ -81,25 +81,25 @@ const Accordion = ({
   title,
   description,
   defaultOpen = false,
-  anchorId,
+  anchorSlug,
 }: AccordionProps) => {
   const [open, setOpen] = useState<boolean>(defaultOpen);
   const titleParts = useMemo(() => parseTitleWithCode(title), [title]);
 
   useLayoutEffect(() => {
-    if (!anchorId) return;
+    if (!anchorSlug) return;
     const syncFromHash = () => {
-      if (getHashFragment() === anchorId) {
+      if (getHashFragment() === anchorSlug) {
         setOpen(true);
       }
     };
     syncFromHash();
     window.addEventListener("hashchange", syncFromHash);
     return () => window.removeEventListener("hashchange", syncFromHash);
-  }, [anchorId]);
+  }, [anchorSlug]);
 
   return (
-    <Box role="listitem" id={anchorId}>
+    <Box role="listitem" id={anchorSlug}>
       <MenuItem
         as="button"
         onClick={() => setOpen(!open)}
--- a/components/ui/Accordion.tsx
+++ b/components/ui/Accordion.tsx
@@ -2,7 +2,7 @@ import { Box, Stack } from "@telegraph/layout";
 import { MenuItem } from "@telegraph/menu";
 import { Icon } from "@telegraph/icon";
 import { AnimatePresence, motion } from "framer-motion";
-import { useState, useMemo } from "react";
+import { useState, useMemo, useLayoutEffect } from "react";
 import { Text, Code } from "@telegraph/typography";
 import { ChevronRight } from "lucide-react";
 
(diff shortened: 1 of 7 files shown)


components/ui/Accordion.tsx
```diff
@@ -15,11 +15,24 @@ const AccordionGroup = ({ children }) => (
...
+  }
+}
+
 type AccordionProps = {
   children: React.ReactNode;
   title: string;
   description?: string;
   defaultOpen?: boolean;
+  /** When set, this value is used as the element `id` and the accordion opens if the URL hash matches (for deep links). */
+  anchorId?: string;
```
scotidodson: I think it could be worthwhile to add a helper to ensure this string is properly formatted for the URL. I know from your example that we intend to supply the hyphenated slug, but could see that being missed during usage. Alternatively, maybe altering the param name to `anchorSlug` helps reinforce the expectation?

content/managing-recipients/setting-channel-data.mdx
```diff
@@ -241,7 +241,7 @@ The `PushDevice` object is used to optionally set device-level metadata for a pu
     | incoming_webhook.url | `string` | The Discord incoming webhook URL (to be used instead of the properties above) |
 
   </Accordion>
-  <Accordion title ="Microsoft Teams">
+  <Accordion title ="Microsoft Teams" anchorId="microsoft-teams-channel-data">
```
scotidodson: While testing the link to this I noticed that we don't totally scroll this open accordion into view. I assume what's happening is we're calculating the height before the accordion is open, and it seems like that might be a common issue with accordion element UX. 

I tested locally and confirmed this concern will mostly be relevant to FAQ sections or the rare non-FAQ accordion at the bottom of the page like this. That doesn't really help except to suggest there may be something hacky to be done to indicate a scroll to bottom.  

If we want to address this, I'd be fine with holding it as a follow up improvement.

scotidodson (approved):
> Should we do a broader audit to add anchorId across the docs, or just add them as we notice they'd be useful? My take is that we don't need it on every accordion but rather only where we have a stable deep link that should open a specific one. If an accordion is unlikely to be linked to directly, a heading anchor is probably enough. But open to other opinions!
I think a broad update would be beneficial so we don't end up doing a bunch of one-off changes, but I don't think the need is urgent. Maybe we could take a step back and see if there's anything else we want to broadly address related to FAQ sections (since those are the most common use case) so we can do that at the same time.  E.g., normalizing writing style of the Qs. 

> Does the approach I've taken with Accordion make the most sense? I'm very open to other opinions here as well!
Makes sense to me, but it'd be great to get a glance from Jeff in case he has any suggestions for handling it differently.

components/ui/Accordion.tsx
rachael-t: I like that suggestion - `anchorSlug` feels more helpful!

content/managing-recipients/setting-channel-data.mdx
rachael-t: Interesting - I noticed it a little bit while working on this (I think because as you mentioned, it's at the bottom of the page) but it feels much worse now (directs you much higher on the page). 🤔

content/multi-tenancy/per-tenant-preferences.mdx
rachael-t, line 1: @scotidodson FYI that I added the `anchorSlug` for the tenant-specific preference merge hierarhcy (since I mentioned it yesterday and for another example to view/test): https://docs-git-rt-kno-7916-linking-to-accordions-knocklabs.vercel.app/multi-tenancy/per-tenant-preferences#tenant-preference-evaluation-rules

rachael-t:
_Finally_ revisited this. Got all the latest changes from main, pulled them locally, and ran a bunch of tests, and this feels good to ship. 🎉 

I updated the PR description to capture the final changes, but one thing to note is that in my testing, I noticed this bug: https://linear.app/knock/issue/KNO-15012/docs-client-side-nav-to-a-different-sidebar-group-lands-mid-page, but this is NOT related to the changes made in this PR and is actually pre-existing/live on the prod docs site and will need to be addressed separately.

PR #1388, opened 2026-04-10
chore(KNO-12588): Add information about test run trigger frequency behavior

rachael-t:
### Description

Adds information to the docs (testing workflows and debugging workflow pages) to clarify that trigger frequency settings are not enforced during test runs. 

https://docs-git-rt-kno-12588-trigger-frequency-callout-knocklabs.vercel.app/send-notifications/testing-workflows#the-workflow-test-runner

https://docs-git-rt-kno-12588-trigger-frequency-callout-knocklabs.vercel.app/send-notifications/debugging-workflows#understanding-workflow-execution-behavior

### Tasks

[KNO-12588](https://linear.app/knock/issue/KNO-12588/docs-add-callout-to-testing-workflows-page-clarifying-that-trigger)

### DIFF
--- a/content/send-notifications/triggering-workflows/overview.mdx
+++ b/content/send-notifications/triggering-workflows/overview.mdx
@@ -46,4 +46,15 @@ When you specify "Once per recipient" frequency, you can include the tenant in t
     [error](/api-reference/overview/errors) when triggered and will not generate
     any workflow recipient runs.
   </Accordion>
+  <Accordion title="Is trigger frequency enforced during test runs?">
+    No. [Trigger
+    frequency](/send-notifications/triggering-workflows/overview#controlling-workflow-trigger-frequency)
+    is not enforced by the [workflow test
+    runner](/send-notifications/testing-workflows). Settings like "once per
+    recipient" are bypassed — the workflow will execute for the selected
+    recipient every time you run a test, regardless of your frequency setting.
+    This is expected behavior. To test trigger frequency as it works in
+    production, trigger the workflow via the [trigger
+    API](/send-notifications/triggering-workflows/api) directly.
+  </Accordion>
 </AccordionGroup>
--- a/content/send-notifications/debugging-workflows.mdx
+++ b/content/send-notifications/debugging-workflows.mdx
@@ -63,6 +63,27 @@ You can also access it from our "Logs" page, which you'll find in the left-hand 
 
 When a workflow runs, each step is executed in sequence. Sometimes an individual step is skipped or encounters an error, and it's important to understand how this affects the workflow run as a whole.
 
+<Callout
+  type="info"
+  title="Note:"
+  text={
+    <>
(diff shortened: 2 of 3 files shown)


cellomatt:
I'd be curious if others have thoughts, but IMO the "some workflow settings" language is a bit confusing here because it implies that there might be other settings in addition to trigger frequency that are ignored but which we don't list here.

I wonder whether a better place to put this would just be as a callout directly on the trigger frequency docs? https://docs.knock.app/send-notifications/triggering-workflows/overview#controlling-workflow-trigger-frequency

That feels like the place where you'd most want to see this.

rachael-t:
@cellomatt yeah that's a good point on the use of "some workflow settings" language and I can adjust if others agree!

I added it to these two pages because this gotcha really only surfaces when using the test runner or looking at the debugger (and it's not obvious from the debugger why it's happening). I considered the trigger frequency page but it's already pretty sparse and I didn't want to add confusion there. Also hoping that having it on these pages also helps our agent surface it when customers are troubleshooting. Curious what others think though!

andy-knock (approved):
@rachael-t LGTM

scotidodson:
> @cellomatt yeah that's a good point on the use of "some workflow settings" language and I can adjust if others agree!
> 
> I added it to these two pages because this gotcha really only surfaces when using the test runner or looking at the debugger (and it's not obvious from the debugger why it's happening). I considered the trigger frequency page but it's already pretty sparse and I didn't want to add confusion there. Also hoping that having it on these pages also helps our agent surface it when customers are troubleshooting. Curious what others think though!

I agree the 'some workflow settings' would create questions for me. In terms of location, I think the two debugging / testing you've selected make the most sense. I do think it could be helpful to add an FAQ to the page Matt linked though for "I have x setting but y is happening???" IMO, a callout would be too much on that page.

content/send-notifications/triggering-workflows/overview.mdx
```diff
@@ -46,4 +46,15 @@ When you specify "Once per recipient" frequency, you can include the tenant in t
...
+    No. [Trigger
+    frequency](/send-notifications/triggering-workflows/overview#controlling-workflow-trigger-frequency)
+    is not enforced by the [workflow test
+    runner](/send-notifications/testing-workflows). Settings like "once per
+    recipient" are bypassed — the workflow will execute for the selected
+    recipient every time you run a test, regardless of your frequency setting.
+    This is expected behavior. To test trigger frequency as it works in
+    production, trigger the workflow via the [trigger
+    API](/send-notifications/triggering-workflows/api) directly.
+  </Accordion>
```
rachael-t, line 59: @cellomatt @scotidodson added this FAQ: https://docs-git-rt-kno-12588-trigger-frequency-callout-knocklabs.vercel.app/send-notifications/triggering-workflows/overview#frequently-asked-questions

LMK your thoughts!

PR #1408, opened 2026-04-23
docs(KNO-12856): point React Native SDK docs at the example app

webbjordy:
### Description

Adds a callout to the React Native SDK Quick start and a link from the SDK overview's Links section, both pointing at the new in-tree example app at `knocklabs/javascript/examples/react-native-example`. Companion to https://github.com/knocklabs/javascript/pull/976.

### Tasks

[KNO-12856](https://linear.app/knock/issue/KNO-12856)


webbjordy:
Merges after [knocklabs/javascript#976](https://github.com/knocklabs/javascript/pull/976) so the example app path exists on main when this lands.

PR #1502, opened 2026-06-23
feat(kno-13896): adds AgentDeeplink component

MikeCarbone:
### Description

Example:
Screenshot 2026-06-23 at 12 47 58 PM

### Usage

Default button with a pre-filled prompt (auto-submits)
```
import { AgentDeeplinkButton } from "@components/ui/AgentDeeplinkButton";

<AgentDeeplinkButton prompt="Create a welcome email workflow for new signups" />
```

Custom label without auto-submit

```
import { AgentDeeplinkButton } from "@components/ui/AgentDeeplinkButton";

<AgentDeeplinkButton
  text="Open in Knock agents"
  prompt="Help me set up a churn prevention workflow"
  autoSubmit={false}
/>
```

Custom styling via Button.Root and Button.Text props
```
import { AgentDeeplinkButton } from "@components/ui/AgentDeeplinkButton";

<AgentDeeplinkButton
  text="Try this workflow"
  prompt="Build a notification workflow for password reset emails"
  mt="6"
  textProps={{ size: "3" }}
/>
```


components/ui/AgentDeeplinkButton.tsx
```diff
@@ -0,0 +1,147 @@
+import React, { useEffect, useRef } from "react";
+import { Button } from "@telegraph/button";
+
+type ButtonRootProps = React.ComponentProps<typeof Button.Root>;
+type ButtonTextProps = React.ComponentProps<typeof Button.Text>;
+
+export interface AgentDeeplinkButtonProps
```
thomaswhyyou, line 7: Does this need to be exported?

components/ui/AgentDeeplinkButton.tsx
```diff
@@ -0,0 +1,146 @@
...
+      });
+
+      frameId = requestAnimationFrame(draw);
+    };
+
+    frameId = requestAnimationFrame(draw);
+
+    return () => {
+      cancelAnimationFrame(frameId);
+    };
```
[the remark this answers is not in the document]
thomaswhyyou: 👀

components/ui/AgentDeeplinkButton.tsx
```diff
@@ -0,0 +1,147 @@
...
+      }}
+    />
+  );
+};
+
+export const AgentDeeplinkButton = ({
+  text = "Try in Knock",
+  prompt,
+  autoSubmit = true,
+  textProps = {},
```
thomaswhyyou, line 112: If this is used for size, then maybe just make it a size prop. Not sure if it needs to be shared across Button.Root and Button.Text

thomaswhyyou (approved):

components/ui/AgentDeeplinkButton.tsx
MikeCarbone: it's not being used for anything right now — just giving us a way to customize the text component if we need to (I can see us wanting to adjust `weight` or `size` depending on where we use the button)

components/ui/AgentDeeplinkButton.tsx
MikeCarbone: I usually like to make component props available for future use in case we need to compose this as part of a larger component to keep props synced `text: AgentDeeplinkButtonProps["text"]`

don't need it right now but also doesn't hurt

components/ui/AgentDeeplinkButton.tsx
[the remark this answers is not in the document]
MikeCarbone: false positive

components/ui/AgentDeeplinkButton.tsx
```diff
@@ -0,0 +1,146 @@
...
+        ctx.roundRect(x, y, CELL_SIZE, CELL_SIZE, 0.75);
+        ctx.fill();
+      });
+
+      frameId = requestAnimationFrame(draw);
+    };
+
+    frameId = requestAnimationFrame(draw);
+
+    return () => {
```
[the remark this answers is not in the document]
MikeCarbone: intentional, this is fine

PR #1506, opened 2026-06-25
docs: document sandbox mode and skip delays for the workflow test runner

cjbell:
### Description

Documents two new settings for the workflow test runner:

- **Sandbox mode** — when enabled, all messages sent from the workflow are forced to send in [sandbox mode](https://docs.knock.app/integrations/overview#sandbox-mode), generating and previewing messages without delivering them to the underlying providers.
- **Skip delays** — when enabled, all [delay steps](https://docs.knock.app/designing-workflows/delay-function) in the workflow are force skipped, so the run executes straight through without waiting.

Both settings are also documented on the trigger API page via the `settings.sandbox_mode` and `settings.skip_delay` properties, which is useful when testing workflows programmatically.

Changes:
- `content/send-notifications/testing-workflows.mdx` — adds a "Test run settings" section.
- `content/send-notifications/triggering-workflows/api.mdx` — adds `settings` to the trigger payload table and a "Controlling trigger behavior with settings" section.

Note: `settings.skip_delay` is not yet present in the OpenAPI spec on `main` (only `sandbox_mode` is), so it won't render in the auto-generated API reference until the spec catches up.

### Tasks

Made with [Cursor](https://cursor.com)

### DIFF
--- a/content/send-notifications/testing-workflows.mdx
+++ b/content/send-notifications/testing-workflows.mdx
@@ -20,6 +20,10 @@ Two things to know about workflow test configuration:
 
 After clicking "Run test," you will see a confirmation and a link to see the log to review the output and what was sent. You can learn more about Knock logs and the debugger [here](/send-notifications/debugging-workflows).
 
+Note that [trigger frequency](/send-notifications/triggering-workflows/overview#controlling-workflow-trigger-frequency) is not enforced during test runs. Settings like "once per recipient" are bypassed — each time you run a test, the workflow will execute for the selected recipient regardless of your frequency setting. To validate trigger frequency as it behaves in production, use the [trigger API](/send-notifications/triggering-workflows/api).
+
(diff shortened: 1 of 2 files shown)


content/send-notifications/triggering-workflows/api.mdx
```diff
@@ -207,6 +208,28 @@ tenants to which they belong in your product.
 
 You can read more about [supporting multi-tenancy in our documentation](/concepts/tenants).
 
+## Controlling trigger behavior with settings
```
rachael-t: Wondering if we should update the title? Unclear if the `settings` object might expand use cases in the future, but for now, since it's just scoped to testing I think we should make that more clear? Open to suggestions on this though!
```suggestion
## Testing workflow triggers
```

content/send-notifications/testing-workflows.mdx
```diff
@@ -20,6 +20,15 @@ Two things to know about workflow test configuration:
 
 After clicking "Run test," you will see a confirmation and a link to see the log to review the output and what was sent. You can learn more about Knock logs and the debugger [here](/send-notifications/debugging-workflows).
 
+### Test run settings
```
rachael-t, line 27: Should we move this to after the source event paragraph and before the callout? It feels a bit disjointed being inserted into this existing section (curious what others think)

PR #1517, opened 2026-07-02
docs: clone resource into environment

emisilvacab:
### Description

Docs about cloning resources across environments, this clarifies some stuff about how to use this feature, like the **production write access** that need to be enabled in order to write to all envs

### DIFF
--- a/content/version-control/environments.mdx
+++ b/content/version-control/environments.mdx
@@ -150,7 +150,7 @@ You can clone the following resources across environments:
 - Partials
 - Translations
 - Reusable requests
-- Source events
+- Source event mappings
 
 Archived resources cannot be cloned.
 
@@ -168,9 +168,9 @@ When you clone a workflow, Knock also copies its channel step templates and work
 Which environments you can clone into depends on your account's **production write access** setting on the **Permissions** page in your account settings:
 
 1. **Production write access enabled.** You can clone into any environment in your project.
-2. **Production write access disabled.** You can only clone into sandbox environments (typically your development environment).
+2. **Production write access disabled.** You can only clone into your Development environment.
 
-This follows the same rules as creating and editing content in an environment. If production write access is disabled, non-sandbox environments such as production and staging are read-only and cannot be selected as a clone destination.
+This follows the same rules as creating and editing content in an environment. If production write access is disabled, environments other than Development are read-only and cannot be selected as a clone destination.
 
 Branch environments cannot be used as clone targets.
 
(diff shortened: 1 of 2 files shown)


content/version-control/environments.mdx
```diff
@@ -163,9 +217,11 @@ There are two tools you can use to control access to your data in the Knock dash
...
-    Yes, but you'll need to move it manually. You can create a new resource in
-    your development environment that uses the same `key` that you used in
-    production. When you promote it to production later on, it will overwrite
-    the existing version that was previously created directly in production.
+    Yes. Use [cloning resources across
+    environments](/version-control/environments#clone-resources-across-environments)
+    to copy the resource into your development environment. You can keep the
+    default `-copy` key if you want a separate copy to experiment with, or set
+    the key to match the production resource if you plan to promote it back and
+    overwrite the production version later.
```
cellomatt: I think we should leave this part of the original:

> When you promote it to production later on, it will overwrite the existing version that was previously created directly in production.

Maybe prior to the sentence about the `-copy` key?

content/version-control/environments.mdx
```diff
@@ -126,6 +129,57 @@ These rules enable two ways of working with content and shipping it to your end
...
+- Workflows
+- Broadcasts
+- Guides
+- Message types
+- Email layouts
+- Audiences
+- Partials
+- Translations
+- Reusable requests
+- Source events
```
cellomatt: Maybe "Source event configurations" or "Source event mappings"? We could leave this the way it is, but there might be an opportunity to be clearer about what is actually being cloned here.

content/version-control/environments.mdx
```diff
@@ -126,6 +129,57 @@ These rules enable two ways of working with content and shipping it to your end
...
+2. **Promote.** Moves committed changes for an existing resource up through your environment chain, keeping the same key. Use this when you are following the development-to-production workflow and want the same resource to exist across environments.
+
+When you clone a workflow, Knock also copies its channel step templates and workflow schema into the destination environment.
+
+### Environment targeting
+
+Which environments you can clone into depends on your account's **production write access** setting on the **Permissions** page in your account settings:
+
+1. **Production write access enabled.** You can clone into any environment in your project.
+2. **Production write access disabled.** You can only clone into sandbox environments (typically your development environment).
```
cellomatt: What is a "sandbox environment," and will it ever be anything other than Development?

content/version-control/environments.mdx
```diff
@@ -126,6 +129,57 @@ These rules enable two ways of working with content and shipping it to your end
...
+
+This follows the same rules as creating and editing content in an environment. If production write access is disabled, non-sandbox environments such as production and staging are read-only and cannot be selected as a clone destination.
```
cellomatt:
```suggestion
This follows the same rules as creating and editing content in an environment. If production write access is disabled, environments other than Development are read-only and cannot be selected as a clone destination.
```
If the answer to the above is that it's only development, I think we should change this too.

content/version-control/environments.mdx
cellomatt:
```suggestion
2. **Production write access disabled.** You can only clone into your Development environment.
```
Here's how I would change it, unless there's another separate sandbox concept that I'm not aware of.

content/version-control/environments.mdx
cellomatt: Or maybe this is simpler:

```suggestion
    Yes. Use [cloning resources across
    environments](/version-control/environments#clone-resources-across-environments)
    to copy the resource into your development environment. 
    
    By default a `-copy` suffix will be added to the key of the cloned resource. Set
    the key to match the production resource that you're cloning 
    if you want promotions from the lower environment to overwrite the production version later on.
```

cellomatt (approved):
Just a few nits, but lgtm!

PR #1540, opened 2026-07-14
chore(kno-14126): clarify warehouse sync backfill timelines

cellomatt:
### Description

Adds explicit mentions of how much historical data is synced for each warehouse table upon initial sync:

- `messages`: Full message history that is available in Knock
- `message_events`: 90 days of historical data
- `recipient_change_stream`: A snapshot of current user state

### DIFF
--- a/content/integrations/extensions/data-sync.mdx
+++ b/content/integrations/extensions/data-sync.mdx
@@ -57,6 +57,8 @@ We currently support the following destination database types:
 
 Our data warehouse connector syncs data from our `messages` table. Each `message` represents a notification that was executed for a single recipient. You can read more about messages as a concept [here](/concepts/messages).
 
+The backfill of messages on initial connection will include all historical message data that is available in Knock.
+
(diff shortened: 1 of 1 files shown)


content/integrations/extensions/data-sync.mdx
```diff
@@ -495,7 +499,7 @@ Each row has an `event_type` indicating how the event was generated. Possible va
 - `recipient.snapshot` - is emitted every time a recipient's properties or preferences are updated, and contains the complete set of properties and preferences at that moment in time. This event may also be generated manually for all or a subset of recipients (by ID) by contacting support.
 - `recipient.deleted` - contains no properties or preferences, but indicates when a recipient was deleted. A recipient may be re-identified after being deleted, which will generate another recipient.created event
 
-Events in the recipient change stream table are retained for 7 days.
+Events in the recipient change stream table are retained for 7 days. The historical backfill on initial connection will include the previous 7 days of data.
```
cellomatt: This may not be accurate, as this table requires a feature flag to be on and I'm not sure we backfill or if we start ingesting events moving forward.

cellomatt: Ok, was able to confirm that we actually only support an initial snapshot here (current state) as opposed to a backfill. Will update this.

thomaswhyyou (approved):

PR #1543, opened 2026-07-15
docs: clarify how tenants hold the Slack access token

meryldakin:
## Summary

The tenant-holds-the-Slack-token story is one of the most confusing parts of our Slack docs: the clearest explanation was buried, unlinked, and worded as a vague "merge." This makes the model explicit and gives it a single canonical home.

- **New "How Slack delivery works" section** on the Slack overview splits a Slack connection into its two halves — **auth** (a bot access token or an incoming webhook) and **destination** (`channel_id`/`user_id`) — and says where each lives.
- **Rewrites the "Tenant channel data requirements" section** to state the actual precedence rule — the tenant's token wins over any token on the recipient (verified against the send-path code, which overwrites the recipient's `access_token` with the tenant's) — instead of the ambiguous "combine." Adds a **"Do you need a tenant?"** note so readers with a single internal workspace aren't over-steered toward tenants.
- **Fixes the cross-referencing gap.** The canonical section had no inbound links from the pages where confused readers start. Adds links from the [tenants concept page](content/multi-tenancy/overview.mdx) (which never mentioned holding credentials), the SlackKit **"About tenants"** and **"Merging channel data"** sections, and the **setting-channel-data** Slack schema.

### DIFF
--- a/content/integrations/chat/slack/overview.mdx
+++ b/content/integrations/chat/slack/overview.mdx
@@ -169,6 +169,15 @@ Here's an example of setting channel data on an `Object` in Knock.
   }
 />
 
+### How Slack delivery works
+
+Every Slack notification needs two things, and Knock lets them live in different places:
+
+- **Auth** — permission to post: either a bot **access token** (`xoxb-…`) from Slack's OAuth flow, or a self-contained **incoming webhook URL**. A bot token authorizes exactly one Slack workspace.
+- **Destination** — where the message goes: a `channel_id` (a channel) or a `user_id` (a direct message).
+
+The **destination** is stored on the recipient (a [`User`](/concepts/users) or [`Object`](/concepts/objects)). The **auth** can be stored on the recipient too — or, in a multi-workspace product, once on the [tenant](#tenant-channel-data-requirements), so every recipient in that tenant shares a single connection. The two sections below cover each location.
+
 ### Recipient channel data requirements
 
 Here's an overview of the data requirements for [setting recipient channel data](/send-notifications/setting-channel-data) for either an incoming webhook or an access token Slack connection. Both will need to live under the `connections` key.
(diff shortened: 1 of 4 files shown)


content/integrations/chat/slack/overview.mdx
```diff
@@ -229,7 +238,9 @@ Here's an overview of the data requirements for [setting recipient channel data]
...
+
+**Do you need a tenant?** Store the token on a tenant when the same recipients need to reach _different Slack workspaces depending on which customer_ a notification is for. You don't need one if you post to a single internal workspace (use an incoming webhook — there's no token to share) or only ever use one workspace's token (store it directly on the recipient). Knock never auto-creates a tenant for you (though SlackKit can, on the fly), so reaching for one is a deliberate choice for the multi-workspace case.
```
cellomatt: > when the same recipients need to reach _different Slack workspaces depending on which customer_ a notification is for.

I'm not sure this is actually the use case for this? Because I don't think a `channel_id` or `user_id` (i.e. the "destination") would remain consistent across Slack workspaces.

In my mind, the main reason to store it on the tenant (besides that this is how our components handle it) is convenience; store once, apply across all notifications in that workspace without needing to persist the `access_token` across recipients. This is especially useful as you're adding additional recipients within the workspace later on

cellomatt: In other words, this isn't necessarily a matter of just swapping out auth but being able to use the recipient's `channel_data` across workspaces as you do that

content/multi-tenancy/overview.mdx
```diff
@@ -14,6 +14,7 @@ You use tenants in Knock to:
 - Apply [per-tenant branding](/multi-tenancy/per-tenant-branding) in emails.
 - Define [per-tenant preference defaults](/multi-tenancy/per-tenant-preferences#create-a-per-tenant-default-preferenceset) that apply to all users within that tenant and [per-user, per-tenant preferences](/multi-tenancy/per-tenant-preferences#set-a-per-tenant-user-preferenceset).
 - Apply [per-tenant translations](/multi-tenancy/per-tenant-translations).
+- Hold [integration credentials](/integrations/chat/slack/overview#tenant-channel-data-requirements), like a Slack workspace's access token, so one connection serves every user in the tenant.
```
cellomatt: I wonder if this should link to something in the channel data concept instead? This PR is a great idea and also makes me realize we should do a similar update for MS Teams, so it might be nice if this mention was slightly-more generic

cellomatt (approved):
Couple small things but otherwise lgtm

meryldakin:
* **#1543** <a href="https://app.graphite.com/github/pr/knocklabs/docs/1543?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/knocklabs/docs/1543?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

content/multi-tenancy/overview.mdx
meryldakin: lifted up to channel data

PR #1587, opened 2026-08-05
docs: documents chat message replies

meryldakin:
## Summary

- document previous-step chat replies as the recommended workflow
- explain external Slack reply identifiers and destination requirements
- add a trigger-data example, troubleshooting guidance, and related-page links

## Validation

- `prettier --check` on changed files
- `yarn lint --quiet`
- `yarn type-check` *(blocked by the existing missing `posthog-js` module/type; reproduced in the untouched checkout)*

### DIFF
--- a/content/designing-workflows/replying-to-chat-messages.mdx
+++ b/content/designing-workflows/replying-to-chat-messages.mdx
@@ -1,0 +1,100 @@
+---
+title: Replying to chat messages
+description: Learn how to reply to messages created by an earlier workflow step or outside Knock.
+tags: ["steps", "channels", "chat", "slack"]
+section: Designing workflows
+---
+
+Chat channel steps can start a new conversation or reply in an existing provider thread. When the parent message was created by the same workflow run, reply to the previous Knock step. Use an external message ID only when the parent message was created outside the workflow.
+
+<Callout
+  type="info"
+  title="Slack support."
+  text="Chat replies are currently available for Slack channel steps."
+/>
+
+## Reply to a previous Knock step
+
+This is the recommended way to create a reply. Knock uses the earlier step's provider message reference and destination, so you do not need to pass provider IDs or configure the destination again.
+
+In the workflow builder:
+
+1. Add a Slack channel step after the step that creates the parent message.
+2. Set **Message behavior** to **Reply to a previous Knock step**.
+3. Select the **Parent step**.
+
+A step is available as a parent when it:
+
+- Is a previous chat channel step that is reachable on every execution path to the reply.
+- Uses the same single Knock channel as the reply step. Channel groups are not supported.
+- Resolves to one provider message and destination for the recipient workflow run.
+
+The reply step inherits the exact Slack message reference and recipient connection from its parent. If the parent message is queued or scheduled, Knock waits for its provider message reference before attempting the reply.
+
+## Reply to an external message
+
+Use an external reply when the parent Slack message was created outside the current workflow. You must provide both the parent message ID and the destination containing that message.
+
+In Slack, the **Parent message ID** is the message timestamp (`ts`) used as `thread_ts` when posting the reply. It commonly comes from workflow trigger data:
+
+```liquid
+{{ data.parent_message_id }}
+```
+
+<Callout
+  type="warning"
+  title="The parent message ID does not select a Slack channel."
+  text="A Slack message timestamp identifies the parent inside a channel. The reply destination must resolve to the same channel, or Slack will not attach it to that thread."
+/>
+
+### Select the destination
+
+For a Slack channel managed through the Knock Slack extension, use **Send to** to set the destination in one of two ways:
+
+- **Select destination.** Choose a channel or person from the connected Slack workspace.
+- **Use channel ID.** Provide a Liquid value that renders to a Slack channel ID, such as `{{ data.channel_id }}`.
+
+The managed Slack connection supplies authentication. Do not put a Slack bot token in workflow data or in the channel ID field.
+
+If you do not configure a destination override, Knock uses the Slack connections stored on the workflow recipient. That is only appropriate when the stored connection points to the same channel as the external parent message.
+
+### Trigger-data example
+
+Send the Slack message timestamp and channel ID in the workflow trigger data:
+
+```javascript
+await knock.workflows.trigger("reply-to-slack-message", {
+  recipients: ["user_123"],
+  data: {
+    parent_message_id: "1784920923.818589",
+    channel_id: "C0123456789",
+  },
+});
+```
+
+Then configure the Slack step with:
+
+| Setting           | Value                          |
+| ----------------- | ------------------------------ |
+| Message behavior  | Reply to an external message   |
+| Parent message ID | `{{ data.parent_message_id }}` |
+| Send to           | Use channel ID                 |
+| Channel ID        | `{{ data.channel_id }}`        |
+
+At execution time, Knock renders the destination as a recipient connection equivalent to:
+
+```json
+{
+  "channel_id": "C0123456789"
+}
+```
+
+## Troubleshooting
+
+| Problem                                                                        | What happens                                                                                            | How to fix it                                                               |
+| ------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------- |
+| The message timestamp is correct, but the channel is omitted                   | Knock uses the recipient's stored Slack connection, which can send to another channel.                  | Select the matching destination or pass its `channel_id`.                   |
+| The message timestamp and channel ID refer to different channels               | The reply does not attach to the expected thread, or Slack rejects the request.                         | Pass the channel that contains the parent message.                          |
+| A Liquid value is empty or malformed                                           | The destination cannot be rendered into a valid recipient connection.                                   | Verify the workflow trigger data and Liquid path.                           |
+| A previous step is not available in the parent-step list                       | The step is not guaranteed to run first, uses another Knock channel, or is inside an unsupported group. | Adjust the workflow graph or use an external reply with explicit IDs.       |
+| A previous step produces no message or more than one possible provider message | Knock cannot identify one parent message and destination safely.                                        | Make the parent resolve to exactly one Slack message for the recipient run. |
--- a/content/integrations/chat/slack/sending-an-internal-message.mdx
+++ b/content/integrations/chat/slack/sending-an-internal-message.mdx
@@ -118,3 +118,5 @@ With that, you should see a message in your selected Slack channel:
   width={2000}
   height={552}
 />
+
+To continue an existing Slack thread instead of starting a new message, see [replying to chat messages](/designing-workflows/replying-to-chat-messages). Replies can inherit the destination from a previous Knock step or use an external Slack message timestamp and matching channel ID.
--- a/data/sidebars/platformSidebar.ts
+++ b/data/sidebars/platformSidebar.ts
@@ -110,6 +110,10 @@ export const PLATFORM_SIDEBAR: SidebarSection[] = [
       },
       { slug: "/step-conditions", title: "Step conditions" },
       { slug: "/channel-step", title: "Channel steps" },
+      {
+        slug: "/replying-to-chat-messages",
+        title: "Replying to chat messages",
+      },
       { slug: "/send-windows", title: "Send windows" },
       { slug: "/validating-trigger-data", title: "Validating trigger data" },
     ],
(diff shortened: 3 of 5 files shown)


data/sidebars/platformSidebar.ts
```diff
@@ -110,6 +110,10 @@ export const PLATFORM_SIDEBAR: SidebarSection[] = [
       },
       { slug: "/step-conditions", title: "Step conditions" },
       { slug: "/channel-step", title: "Channel steps" },
+      {
+        slug: "/replying-to-chat-messages",
+        title: "Replying to chat messages",
+      },
```
cellomatt: I think it might be better to have this page under the "Chat" section of our integrations docs? Maybe in between Overview and Slack. We use a similar pattern for channel-specific things like email attachments, for example (i.e. things that are configured on the specific channel step type). 

Image

cellomatt: (noting that we'll also need to update any hyperlinks from other pages above if we change the page's location)

content/integrations/chat/slack/sending-an-internal-message.mdx
```diff
@@ -110,6 +110,8 @@ await knock.workflows.trigger("support-alert", {
 
 The workflow recipient still provides the `recipient` variables used in Liquid templates, step conditions, and the rest of the workflow. The destination selected in **Send to** only changes where that internal Slack step is delivered.
 
+To continue an existing Slack thread instead of starting a new message, see [replying to chat messages](/designing-workflows/replying-to-chat-messages). Replies can inherit the destination from a previous Knock step or target an existing thread using its message timestamp and matching channel ID.
```
cellomatt:
```suggestion
To continue an existing Slack thread instead of starting a new message, see [replying to chat messages](/designing-workflows/replying-to-chat-messages). Replies can inherit the destination from a previous workflow step or target an existing thread using its message timestamp and matching channel ID.
```

content/designing-workflows/replying-to-chat-messages.mdx
```diff
@@ -0,0 +1,96 @@
+---
+title: Replying to chat messages
+description: Learn how to reply to messages created by an earlier workflow step or in an existing thread.
+tags: ["steps", "channels", "chat", "slack", "discord"]
+section: Designing workflows
```
cellomatt: Will need to be updated if we move

content/designing-workflows/replying-to-chat-messages.mdx
```diff
@@ -0,0 +1,96 @@
...
+---
+
+Chat channel steps can start a new conversation or reply in an existing provider thread. When the parent message was created by the same workflow run, reply to the previous Knock step. When the parent message already exists outside the workflow, provide its message ID instead.
```
cellomatt:
```suggestion
Chat channel steps can start a new conversation or reply to an existing provider thread. When a parent message was created by the current workflow run, you can send a reply to a previous chat step. To reply to a message that already exists outside the workflow run, you'll provide its message ID instead.
```

content/designing-workflows/replying-to-chat-messages.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+<Callout
+  type="info"
+  title="Slack and Discord support"
```
cellomatt:
```suggestion
  title="Supported providers."
```
Just a note that if the copy change isn't accepted, we do need to at a minimum add a `.` for style guide compliance.

content/designing-workflows/replying-to-chat-messages.mdx
```diff
@@ -0,0 +1,96 @@
...
+  text="Chat replies are available for Slack and Discord channel steps that use a bot-based connection. Incoming webhook connections do not support replies."
+/>
+
+## Reply to a previous Knock step
+
+This is the recommended way to create a reply. Knock uses the earlier step's provider message reference and destination, so you do not need to pass provider IDs or configure the destination again.
```
cellomatt:
```suggestion
This is the recommended way to create a reply. Knock uses the earlier workflow step's provider message reference and destination, so you do not need to pass provider IDs or configure the destination again.
```

content/designing-workflows/replying-to-chat-messages.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+In the workflow builder:
+
+1. Add a chat channel step after the step that creates the parent message.
+2. Set **Message behavior** to **Reply to a previous Knock step**.
+3. Select the **Parent step**.
+
+A step is available as a parent when it:
+
+- Is a previous chat channel step that is reachable on every execution path to the reply.
```
cellomatt:
```suggestion
- Is a previous chat channel step that is reachable on every execution path to the reply. Chat steps inside conditional [branches](/designing-workflows/branch-function) will not be eligible as a parent to subsequent steps outside of that branch.
```

content/designing-workflows/replying-to-chat-messages.mdx
```diff
@@ -0,0 +1,96 @@
...
+- Uses the same single Knock channel as the reply step. Channel groups are not supported.
+- Resolves to one provider message and destination for the recipient workflow run.
```
cellomatt:
```suggestion
- Resolves to one provider message and destination for the current recipient workflow run.
```
Question: what does this exclude? Something like a previous chat step that sends multiple messages because the recipient's channel data is an array of channel connections? Should we say that explicitly?

content/designing-workflows/replying-to-chat-messages.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+The reply step inherits the exact provider message reference and recipient connection from its parent. If the parent message is queued or scheduled, Knock waits for its provider message reference before attempting the reply.
+
+## Reply to an existing thread
+
+Use this option when the parent message already exists outside the current workflow.
```
cellomatt:
```suggestion
Use this option when the parent message already exists outside of the current workflow run.
```

content/designing-workflows/replying-to-chat-messages.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+In the workflow builder:
+
+1. Set **Message behavior** to **Reply to an existing thread**.
+2. Enter the **Parent message ID**, usually as a Liquid value from your workflow data.
```
cellomatt:
```suggestion
2. Enter the **Parent message ID**, usually as a dynamic Liquid reference from your workflow trigger `data`.
```

content/designing-workflows/replying-to-chat-messages.mdx
```diff
@@ -0,0 +1,96 @@
...
+3. For Slack, optionally enter the channel ID in **Send to**.
```
cellomatt:
```suggestion
3. For Slack, optionally enter the channel ID in **Send to**. Required if the parent message did not originate in Knock. The value must match the channel ID of the parent.
```
This is required if the parent message was not generated through Knock, correct? What are the scenarios where this wouldn't be necessary? Feel free to tweak suggested language, this is based on my understanding.

content/designing-workflows/replying-to-chat-messages.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+The value expected for **Parent message ID** depends on the provider:
+
+- For Slack, use the parent message timestamp (`ts`). Knock sends it to Slack as `thread_ts`.
```
cellomatt:
```suggestion
- For Slack, use the parent message's timestamp (`ts`). Knock sends it to Slack as `thread_ts`.
```

content/designing-workflows/replying-to-chat-messages.mdx
cellomatt: Ah ok, I see this is covered below. So maybe instead....

```suggestion
3. For Slack, optionally enter the channel ID in **Send to**. See below for more detail on [configuring the destination](#configure-the-destination).
```

Just whatever we can do to mitigate potential confusion about why this is optional and point people to the answer

content/designing-workflows/replying-to-chat-messages.mdx
```diff
@@ -0,0 +1,96 @@
...
+- For Discord, use the parent Discord message ID.
+
+### Configure the destination
+
+For Slack, **Send to** accepts a Slack channel ID, usually from workflow data such as `{{ data.channel_id }}`. Leave it blank to use the recipient's Slack channel connection. In either case, the destination must be the channel that contains the parent message.
```
cellomatt:
```suggestion
For Slack, **Send to** accepts a Slack channel ID, usually from workflow trigger data such as `{{ data.channel_id }}`. Leave it blank to use the recipient's Slack channel connection. In either case, the destination must be the channel that contains the parent message.
```

content/designing-workflows/replying-to-chat-messages.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+For Discord, Knock uses the Discord channel data stored on the recipient. That connection must identify the channel that contains the parent message.
+
+The configured Slack or Discord channel supplies authentication. Do not put a bot token in workflow data or in the destination field.
+
+### Slack trigger-data example
```
cellomatt:
```suggestion
### Slack trigger data example
```

content/designing-workflows/replying-to-chat-messages.mdx
cellomatt: I think we should also note that for a Slack message that was generated by Knock, this `ts` is the same value as the `metadata.external_id` that we return via API/webhooks for that message. We could mention that here, or maybe in a callout on the example below? I think it's a useful implementation pattern to call out

cellomatt: (and I guess the same bot token caveat applies for that use case and should potentially be noted but maybe the mention above covers that well enough already)

content/designing-workflows/replying-to-chat-messages.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+Send the Slack message timestamp and channel ID in the workflow trigger data:
+
+```javascript
```
cellomatt:
```suggestion
```javascript title="Triggering a threaded Slack response"
```
nit: should include a `title` of some sort

content/designing-workflows/replying-to-chat-messages.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+| Setting           | Value                          |
+| ----------------- | ------------------------------ |
+| Message behavior  | Reply to an existing thread    |
+| Parent message ID | `{{ data.parent_message_id }}` |
+| Send to           | `{{ data.channel_id }}`        |
+
+At execution time, Knock renders the destination as a recipient connection equivalent to:
+
+```json
```
cellomatt:
```suggestion
```json title="A recipient Slack connection"
```
Same nit as above

content/designing-workflows/replying-to-chat-messages.mdx
```diff
@@ -0,0 +1,96 @@
...
+  "channel_id": "C0123456789"
+}
+```
+
+## Troubleshooting
+
+| Problem                                                                        | What happens                                                                                            | How to fix it                                                                 |
+| ------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------- |
+| The Slack timestamp is correct, but **Send to** is blank                       | Knock uses the recipient's stored Slack connection, which might point to another channel.               | Enter the channel ID that contains the parent message.                        |
+| The Slack timestamp and channel ID refer to different channels                 | The reply does not attach to the expected thread, or Slack rejects the request.                         | Enter the channel that contains the parent message.                           |
```
cellomatt:
```suggestion
| The Slack timestamp and channel ID refer to different channels                 | The reply does not attach to the expected thread, or Slack rejects the request.                         | Enter the channel ID that contains the parent message.                           |
```
For consistency

content/designing-workflows/replying-to-chat-messages.mdx
```diff
@@ -0,0 +1,96 @@
...
+| The recipient uses an incoming webhook connection                              | Knock cannot reply because incoming webhooks do not provide the required bot-based connection.          | Configure channel data that uses a bot token and channel ID.                  |
+| A Liquid value is empty or malformed                                           | Knock cannot render a valid parent message ID or destination.                                           | Verify the workflow trigger data and Liquid path.                             |
```
cellomatt:
```suggestion
| A Liquid value is empty or malformed                                           | Knock cannot render a valid parent message ID or destination, so no reply is sent.                                           | Verify the workflow trigger data and Liquid path.                             |
```

cellomatt (approved):
Approved, with a bunch of mostly-small nits. 

Main thing that we should change in my opinion is moving this documentation under Integrations > Chat, which will also require updating all of the links that reference the new page.

PR #1593, opened 2026-08-07
chore(kno-9513): batch function clarifications

cellomatt:
### Description

This PR makes several clarifications to the batch function documentation:

- Documents expected behavior of batching by a workflow's version, including the implications of making updates to already-published workflows with open batches
- In light of those implications, adds a callout about batches vs. digests, with a link to the tutorial on powering recurring digests
- Documents relative batch windows (previously missing)
- Adds an FAQ about using multiple batch steps in a single workflow (don't do this sequentially, and use a batch key when used in parallel)

https://docs-ciaikk75f-knocklabs.vercel.app/designing-workflows/batch-function

### DIFF
--- a/content/designing-workflows/batch-function.mdx
+++ b/content/designing-workflows/batch-function.mdx
@@ -351,6 +351,7 @@ To prevent a workflow from sending notifications or accumulating additional even
 
 - A workflow's status is an environment-specific setting. Be sure to target your production environment if you decide to use this control.
 - Toggling a worflow's status does not automatically cancel any currently-open batch windows. Your workflow should remain in the `Inactive` state until all open batch windows have closed, otherwise your batches will proceed as normal once their window durations have elapsed.
+- While the workflow is `Inactive` in Production, promote your changes. After your open batch windows close and outstanding workflows have terminated, setting the workflow back to `Active` will begin batching incoming events again and these workflow runs will proceed with the newly-published workflow version.
 
 ## Using workflow cancellation with batches
 
--- a/content/designing-workflows/batch-function.mdx
+++ b/content/designing-workflows/batch-function.mdx
@@ -10,31 +10,43 @@ A batch function collects notifications that have to do with the same subject, s
 Batch functions are helpful when a recipient needs to be notified about a lot of activity happening at once, but doesn't need a notification for every single activity within the batch.
 Commenting is a common use case. If a user leaves ten comments in a page in fifteen minutes, you don't want to send the user ten separate notifications. You want to send them one notification about the ten comments they just received.
 
+<Callout
+  type="warning"
+  title="Batches vs. digests."
+  text={
+    <>
+      The batch function is a powerful tool for aggregating multiple similar
+      events into a single notification. However, you may find that its inherent
+      characteristics mean that it isn't well-suited for powering regular daily
+      or weekly digests that summarize a user's account activity, where
+      notifications should always send at specific intervals and you need to
+      have granular control over the events that are included in those messages.
+      <br />
+      <br />
+      See our tutorial on <a href="/tutorials/building-recurring-digests">
+        powering recurring digests with Knock
+      </a> for more guidance on the pattern that we recommend. If you aren't sure
+      about the best approach for your use case, please <a href="mailto:support@knock.app?subject=Building recurring digests with Knock">
+        reach out to our support team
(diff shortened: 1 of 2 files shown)


content/designing-workflows/batch-function.mdx
```diff
@@ -10,31 +10,43 @@ A batch function collects notifications that have to do with the same subject, s
...
-  title="Enterprise plan feature."
-  text={
-    <>
-      By default the batch will only return the first (or last) 10 items to be
-      rendered in your template. This limit can be configured on our Enterprise
-      plan (up to 100 items). The batch can, however, accumulate{" "}
-      <strong>any number of items</strong> over the window that it's open.
-    </>
-  }
-/>
```
cellomatt, line 33: I cut this callout because we also have a similar one further down the page where we actually discuss rendering items

content/designing-workflows/batch-function.mdx
```diff
@@ -10,31 +10,43 @@ A batch function collects notifications that have to do with the same subject, s
...
+A batch function always captures events per `recipient` and `workflow_version_id`. For more information on how updates to your workflow will affect existing open batches, see the section on [updating workflows with batch steps](#updating-workflows-with-batch-steps) below.
 
 ## Selecting a batch key
 
-A batch function always captures events per `recipient`. When you provide a batch key, batched events are further grouped by that key. The batch key resolves to a value in your `data` payload. You can also use Liquid to [reference variables](/template-editor/variables) from the workflow run scope (such as the `recipient.*`, `actor.*`, or `tenant.*` namespaces) to construct a key.
+When you configure an optional batch key, batched events are further grouped by that key. The batch key resolves to a value in your `data` payload; a configured batch key of `event_type` points to `data.event_type`. You can also use Liquid to reference [variables](/template-editor/variables) from the workflow run scope (such as the `recipient.*`, `actor.*`, or `tenant.*` namespaces) to construct a key.
```
cellomatt, line 49: Moved the first bit (along with workflow version clarification) up into the previous paragraph.

content/designing-workflows/batch-function.mdx
```diff
@@ -79,7 +92,7 @@ The batch window determines the length of time that the batch will be open, with
 
 ### Set a fixed batch window
 
-You can set a fixed duration batch window using the "Batch for a fixed window" option in the batch step. The window accepts a relative duration, which can be specified in seconds, minutes, hours, or days.
+You can set a fixed duration batch window using the "Fixed" window option in the batch step. The window accepts a duration which can be specified in seconds, minutes, hours, or days.
```
cellomatt, line 95: I cut "relative" here to disambiguate from relative windows

content/designing-workflows/batch-function.mdx
```diff
@@ -320,6 +339,19 @@ When a workflow includes a batch step, the `data` payload that is available to y
...
+
+To prevent a workflow from sending notifications or accumulating additional events while you work on an update, you can set the workflow's [status](/concepts/workflows#workflow-status) to `Inactive`. While a workflow remains in the inactive state, new events will not be batched and any currently-open batch windows that close will not proceed to the next workflow step; their runs will instead be terminated and no additional messages will be generated.
+
+- A workflow's status is an environment-specific setting. Be sure to target your production environment if you decide to use this control.
+- Toggling a worflow's status does not automatically cancel any currently-open batch windows. Your workflow should remain in the `Inactive` state until all open batch windows have closed, otherwise your batches will proceed as normal once their window durations have elapsed.
```
rachael-t, line 353: I think the only piece of feedback I have is that in reading this, it's not explicitly clear when we recommend toggling a workflow's status back to active if someone is following this toggle workaround. Wonder if there is a way to adjust this to make that a bit clearer for customers to follow?

cellomatt: I could add an additional bullet:

- While the workflow is `Inactive` in Production, promote your changes. After your open batch windows close and outstanding workflows have terminated, setting the workflow back to `Active` will begin batching new incoming events again and these workflow runs will proceed with the newly-published workflow version. 

wdyt?

rachael-t: Yeah I think that sounds great and would be helpful to clarify!

rachael-t (approved):

